Issue #42, opened 2021-11-02
Improve LaTeX builder


veli-m-andirin:
In this issue we aim to have a working LaTeX builder that can also handle bibliographies and cross-references.

rcalvo12:
@jmshapir, I just wrapped up my first pass at the improved latex builder (!!!!!!!!) but need write access to share my results.

jmshapir:
> @jmshapir, I just wrapped up my first pass at the improved latex builder (!!!!!!!!) but need write access to share my results.

@rcalvo12 done, thanks!

rcalvo12:
In 402c380 I produced my first pass at the improved latex builder. I appreciate everyone's patience on this. Changes were made in branch `issue42_latex_builder`.

The previous latex builder actually wasn't too far off, and after adjusting some fields in the latex file I could produce a pdf from a latex file that included external tables and graphs, footnotes, and a handwritten reference section. The main changes here were to make sure that the files referenced within the latex file were referenced from root of the directory. For example, instead of calling a table with `\input{../output/tables/top_gdp}` you call it with `\input{output/tables/top_gdp}`. You can see examples in the `TemplateTest.tex` file. 

However, with the previous version of the builder, we could not use a .bib file to produce a bibliography.

The updated builder uses the following sequence to compile the latex files:
```
pdflatex -> bibtex -> pdflatex -> pdflatex
```
If you go to the `/issue42/` directory, delete everything except `References.bib`, `TemplateTest.tex`, and `Sconscript`, and then run `scons` from the root of the repository you can see that it produces the sample pdf with the proper references.

Looking forward to iterating on next versions.

jmshapir:
Thanks @rcalvo12 and nice work! I was able to compile the issue folder successfully on my first try. :-)

A few things:
* I think we should add Template.tex to /souce/paper. (That way users of JMSLab/Template have an easy choice between TeX and LyX.)
* I think we should add References.bib to /source/paper, and we should revise Template.lyx to use BibTeX for references. (I'm inclined to move the lab towards using BibTeX more universally.)
* I think we should update `test_build_latex.py` to test the current instantiation of the builder.

From my standpoint, once we've done those things, you can open a pull request with @veli-m-andirin @jmshapir as reviewers. (At that point we can also see if @mcaceresb wants to review.)

rcalvo12:
> * I think we should add Template.tex to /souce/paper. (That way users of JMSLab/Template have an easy choice between TeX and LyX.)
[...]

Issue #53, opened 2022-03-24
Issue to acknowledge use of Template


jmshapir:
If you are planning to use this template in your own research outside of JMSLab, please post a comment here and acknowledge that you are willing to help participate in Template development.

marcopetterson:
Thank you @jmshapir! I plan to use the template in future research and happy to help with development!

rcalvo12:
Thanks @jmshapir, I plan to use this template in future research and I am happy to contribute to future Template development!

yu-jia-wang:
Hi @jmshapir, I plan to use this template for future research. Happy to help with development!

vshivram:
Hi @jmshapir, thanks for putting this together! I am incorporating this template into my workflow and am happy to contribute however it's helpful.

MaxMillerLab:
@jmshapir, I plan to use this in future work too and am happy to help in development, upkeep, etc.

ben-workman:
@jmshapir I plan on using this template in my own research and agree to help participate in Template development.

mariapolyakova:
hi @jmshapir, thanks so much for making this template public. I'd like to use some parts of the template for my own work. Of course happy to help maintain it. Thanks!

liaochris:
@jmshapir Thanks for open sourcing the template; I plan to use this in my work and happy to contribute to development.

AssafKott:
Hi @jmshapir! I am planning to use the template and more than happy to contribute to development.

julienberman:
Hi @jmshapir, I plan to use this template for my independent research. Am happy to contribute to development.

himalbw:
@jmshapir I plan to use this template for my thesis and am happy to contribute to development; thank you!

PR #56, opened 2022-04-01
Pull Request for #42: Improve LaTeX builder

rcalvo12:
@jmshapir and @veli-m-andirin could you review this PR when you get a chance?

In this issue, we made changes to improve the LaTeX builder so that it could produce papers from LaTeX files that use BibTeX for references. 

Changes:
- Edits to `build_latex.py`
- Addition of `Template.tex` and `References.bib` to source/paper
- Addition of `gdp_educ.tex` and `top_gdp.tex` to be referenced by Template
- Edits `Template.lyx` to use `References.bib` for references
- Edits to `source/paper/sconscript` 

For this PR, I think it would work if:
- [ ] @veli-m-andirin and @jmshapir review the code and repo changes
- [ ] Someone (@veli-m-andirin ?) double checks that `test_build_latex.py` does not need edits
- [ ] @rcalvo12 removes the issue folder and closes the pull

@mcaceresb, if you want to review as well, I can add you as a reviewer.


mcaceresb:
@rcalvo12 Yes, thanks! Will have a look later today.

mcaceresb:
@rcalvo12 Did you run the tests locally? This doesn't seem to be specific to Linux; the tests as they are ought to fail generally. Just want to check this is the case.

Tagging @jmshapir since I think some of these overlap with how this builder ought to be designed. Some wrinkles I've found so far:

- The BibTeX call is evaluated with the same tests as the LaTeX call. The LaTeX tests expect the command-line call to be formatted differently than BibTeX.
- There are no BibTeX-specific tests. I see there was a short back and forth in #42 about this, but I'd argue the existing tests are not testing the BibTeX addition. I think we might want to check that the  function works: Without a `.bib` file, with a `.bib` file, that BibTeX is run on the right target (i.e. build and not source), and that `.bbl`/`.blg` files are created if there is a `.bib` file as input.
- There is no BibTeX-specific executable. This is somewhat of a corner case so it may not be necessary (it would only serve to allow users to specify a custom executable, like a custom path or `biber`).
- The tests are designed so multiple calls to `subprocess.check_output` will fail. I believe the idea behind is is to make sure the builder is invoked only once, so that we can see the tests being run correspond to a single call to the builder. However, in this case the builder makes 4 calls. It should be possible to make multi-line calls but this may not be the best design.

I'll make code-specific comments (if there are any)  in my review. LMK if I should help with any of the above before then.

veli-m-andirin (commented):
Thanks @rcalvo12 @mcaceresb!

Additional high-level thoughts before reviewing the code in detail:

- Currently, it seems that SCons will exit with an error message if the source file does not have a bibliography associated. This is related to @mcaceresb's comment above. I think we want the tests to cover cases with and without bibliography. (And to have passing tests, we'd need to handle these cases correctly, too.)
- I wonder how we'd like to deal with cases where we have additional files that have references to the main text, such as an online Appendix. In that case, In these cases the order in which we compile the files would matter. 
- I think we want a proper clean-up method so that auxiliary files are not committed to the repo. 
- I also wonder if we should implement a clean-up method before making the system calls, too. I recall having to clean auxiliary files after a failed recompilation before attempting to recompile again, so this might be useful, but happy to defer to @mcaceresb on that. 
- @mcaceresb, in case useful, tests for LaTeX builder fail on Windows as well. In addition to what you said, in case we attempt to have a builder that can deal with additional files such as appendices, I think we'd need tests for those, too.

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -44,3 +46,40 @@ def add_call_args(self):
...
+        Raise an informative exception on error.
+        '''
+        self.bibtex_executable  = 'bibtex'
+        self.bibtex_system_call = '%s %s' % (self.bibtex_executable, self.bib_file)
+        
+        traceback = ''
+        raise_system_call_exception = False
+        try:
+            subprocess.check_output(self.system_call, shell = True, stderr = subprocess.STDOUT)
+            subprocess.check_output(self.bibtex_system_call, shell = True, stderr = subprocess.STDOUT)
```
veli-m-andirin, line 71: I think we should have a conditional logic that would make the bibtex call only when a bibliography is used.

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -44,3 +46,40 @@ def add_call_args(self):
...
+            subprocess.check_output(self.system_call, shell = True, stderr = subprocess.STDOUT)
+            subprocess.check_output(self.system_call, shell = True, stderr = subprocess.STDOUT)
+        except subprocess.CalledProcessError as ex:
+            traceback = ex.output
+            raise_system_call_exception = True
+
+        self.cleanup()
```
veli-m-andirin, line 78: It would be good to have a proper cleanup method for latex builder so that auxiliary files are cleaned after (or maybe also before) compiling.

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -44,3 +46,40 @@ def add_call_args(self):
...
+            target_file = ''
+        self.bib_file = target_file
+        return None
+    
+    def do_call(self):
+        '''
+        Acutally execute the system call attribute.
+        Raise an informative exception on error.
+        '''
+        self.bibtex_executable  = 'bibtex'
```
veli-m-andirin, line 64: @mcaceresb, if `biblatex` is used, then should the command be `biber`, and do you think this is something that we should try to allow for?

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -44,3 +46,40 @@ def add_call_args(self):
                                   os.path.normpath(self.log_file))
         self.call_args = args
         return None
+     
+    def add_bib_name(self, target):
+        if bool(target):
+            target = misc.make_list_if_string(target)
+            target_file = str(target[0]).split(".")[0]
```
veli-m-andirin, line 53: I think this wouldn't work if a file name contains multiple dots.

source/paper/sconscript.log
```diff
@@ -0,0 +1,119 @@
+*** Builder log created: {2022-03-31 11:32:00}
```
veli-m-andirin, line 1: I think sconscript.log files should go to the output folder.

veli-m-andirin (commented):
Thanks again @rcalvo12. I left some in-line comments, some of which are related to the points in my previous comment.

source/lib/JMSLab/builders/build_latex.py
mcaceresb: @veli-m-andirin I think it would be useful. However, I don't think we should be too smart about it. It would be sufficient to allow the user to specify a BibTeX executable (like any of the other executables).

source/lib/JMSLab/builders/build_latex.py
mcaceresb: @veli-m-andirin This would not be sufficient as `bibtex` also fails when no citations are defined in the main text (even if there is a bibliography file referenced).

source/lib/JMSLab/builders/build_latex.py
mcaceresb: Agree; we can define a dictionary of file extensions to clean up after LaTeX/BibTeX have run, and delete any files with that extension if their name corresponds to the name of the root file. We should only keep `.log` and `.blg`, I think?

mcaceresb (changes requested):
Thanks @rcalvo12! Made some specific comments to the code. We can iterate if needed.

source/lib/JMSLab/builders/build_latex.py
veli-m-andirin: I think a structure where

1) bibtex is not called when a bibliography is not referenced
2) bibtex is called when a bibliography is referenced

would be fine. What you say would fall into the second case. Even though it would raise an error if citations are absent, I think the solution there would be to modify the source file. (If someone references a bibliography, then it is up to them to use it, and if they don't use it, they can just remove it.) I can also imagine a case where we also confirm that citations are indeed made, but this might be more than needed.

SConstruct
```diff
@@ -24,3 +25,5 @@ SConscript('source/derived/SConscript')
 SConscript('source/analysis/SConscript')
 SConscript('source/tables/SConscript')
 SConscript('source/paper/SConscript')
+SConscript('issue42/SConscript')
```
jmshapir, line 28: Need to remember to remove this prior to merge.

output/analysis/plots/sconscript.log
```diff
@@ -1,4 +1,11 @@
-*** Builder log created: {2021-05-30 10:42:17}
-*** Builder log completed: {2021-05-30 10:42:31}
-[Warning: MATLAB has disabled some advanced graphics rendering features by switching to software OpenGL. For more information, click <a
-href="matlab:opengl('problems')">here</a>.] 
+*** Builder log created: {2022-03-30 13:49:44}
+*** Builder log completed: {2022-03-30 13:50:05}
+[Warning: Function plot has the same name as a MATLAB builtin. We suggest you rename the
```
jmshapir, line 3: Maybe we should rename the function to avoid this warning going forward?

output/paper/TemplateTex.bbl
```diff
@@ -0,0 +1,13 @@
+\begin{thebibliography}{}
```
jmshapir, line 1: Maybe we want to ignore BBL files via the global .gitignore?

output/paper/TemplateTex.blg
```diff
@@ -0,0 +1,46 @@
+This is BibTeX, Version 0.99d (TeX Live 2021)
```
jmshapir, line 1: Suggest to add .BLG to .gitignore

output/paper/TemplateTex.log
```diff
@@ -0,0 +1,393 @@
+This is pdfTeX, Version 3.141592653-2.6-1.40.22 (TeX Live 2021) (preloaded format=pdflatex 2022.2.4)  31 MAR 2022 11:37
```
jmshapir, line 1: Maybe create a local .gitignore that ignores TemplateTex.log? (Since we don't want to ignore .LOG files universally.)

jmshapir (changes requested):
Nice work @rcalvo12! I took a first pass here and will keep an eye on the pull as we address other points.

output/paper/TemplateTex.blg
mcaceresb: @jmshapir Just wanted to highlight `.blg` is the BibTeX log file, in case it makes sense to keep the bibliography log.

jmshapir: Thanks @mcaceresb! For now my instinct is to ignore the various LaTeX/BibTeX-generated logs and keep the sconscript.log. Users can still have a local copy of LaTeX/BibTeX logs for debugging.

rcalvo12: @jmshapir Added to gitignore in 493caac.

output/analysis/plots/sconscript.log
rcalvo12: @jmshapir Done in 0d23e83, error no longer occurs.

output/paper/TemplateTex.log
rcalvo12: @jmshapir The problem here is that in /output/paper there is both TemplateTex.log and sconscript.log, so adding a local .gitignore for '*.log' would also ignore our sconscript log file. We could just have the .gitignore ignore specifically 'TemplateTex.log' but this really wouldn't do anything for a future user of the template since they will certainly change the name of the output files. If we would still prefer to do this for the cleanliness of the Template repo, we can.

jmshapir: @rcalvo12 can we add a local ignore for .LOG but exempt sconscript.log?

rcalvo12: @jmshapir apologies, on my first search I didn't think it was possible, but I believe the change in 63dda7c should do the trick.

jmshapir: @rcalvo12 I agree, that looks good, thanks!

issue42/SConscript
```diff
@@ -0,0 +1,8 @@
+Import('*')
+
+target = ['#issue42/TemplateTest.pdf']
+source = ['#issue42/TemplateTest.tex',
```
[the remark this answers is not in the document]
rcalvo12: @mcaceresb added in 9df167a.

source/lib/JMSLab/builders/build_latex.py
rcalvo12: @veli-m-andirin @mcaceresb In 0ddc2c6 I modified the builder so that it only calls bibtex if a .bib file is included in the sources in scons. If there is a .bib file in sources, it runs the normal pdflatex -> bibtex -> pdflatex -> pdflatex. If there is no .bib file, it runs pdflatex twice as per [this comment](https://github.com/JMSLab/Template/issues/42#issuecomment-961205623). 

@jmshapir fyi.

mcaceresb: @rcalvo12 thanks! Made some minor comments.

As for running it twice, per the comment, 3 is safer, though admittedly it's a bit of an edge case. I've added a simple example in 077b56c5e5cbae266d803a24834be67e568a06d7 where running `pdflatex` twice does not give the right reference.

@veli-m-andirin fyi in case you have thoughts on this point.

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -44,3 +46,72 @@ def add_call_args(self):
...
+    def check_bib(self, source):
+    
+        bibext = '.bib'
+        bib_file = ''
+        
+        if bool(source):
+            sources = misc.make_list_if_string(source)
+            for i in range(len(source)):
+                source_file = str(source[i])
+                if bibext in source_file:
```
mcaceresb, line 67: This matches `.bib` anywhere in `source_file`. It also might be prudent to do the check in lowercase, e.g. `source_file.lower().endswith(bibext)`

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -44,3 +46,72 @@ def add_call_args(self):
...
+                    bib_file = source_file
+        else:
+            bib_file = ''  
+        self.check_bib = bib_file
+        return None
```
mcaceresb, line 72: Minor stuff (not essential):

- To avoid confusion in the future, it might be a good idea to make `self.check_bib` a boolean rather than a file name (it should be possible to have multiple `.bib` files; if `check_bib` is a boolean then it's clear it's not the name of the `.bib` files but a T/F indicator).
- Just as an fyi you can do `for source_file in source:` here.

source/lib/JMSLab/builders/build_latex.py
mcaceresb: @rcalvo12 Thanks! Was it important to have the `target = misc.make_list_if_string(target)` line? Unsure of the answer but flagging that this also took that out. Otherwise looks good.

source/lib/JMSLab/builders/build_latex.py
veli-m-andirin: @mcaceresb, if you don't think there are any serious drawbacks to running three times (I don't think runtime qualifies as a serious drawback), then I definitely agree with doing what is safer.

rcalvo12: @mcaceresb @veli-m-andirin I implemented a third run in 7bfca71, forgot to leave a note here about it.

source/lib/JMSLab/builders/build_latex.py
rcalvo12: @mcaceresb and @veli-m-andirin, I added a cleanup function in 92242a4 which I cleaned up in 9f5d9e6. It cleans up the .aux file and the .bbl file after compiling and leaves behind the log files. Let me know what you think of these edits.

mcaceresb: @veli-m-andirin What's your take on which extensions to include? This is what I have in my own local cleanup file (excluding package-specific extensions). I know many of these probably won't be necessary for the template, though. Should we just include "core" LaTeX (plus `bibtex` and `beamer`)?

```
.aux
.bbl
.bcf
.blg
.brf
.bst
.clo
.cls
.drv
.dtx
.fd
.fls
.glg
.glo
.gls
.glsdefs
.idx
.ilg
.ind
.ins
.ist
.lof
.log
.lot
.nav
.out
.snm
.tikz
.toc
.vrb
```

rcalvo12: @mcaceresb Just to make sure I am on the same page, you are thinking we would add some or all of these to the list started in line 74 of `build_latex.py`, correct?

rcalvo12: @mcaceresb Ok, I added these other filetypes (except for .blg since its a log file) to the list of extensions being cleaned up in 40649c3.

source/lib/JMSLab/builders/build_latex.py
rcalvo12: @veli-m-andirin @mcaceresb In fa4dad5 I added some functionality so that `self.bibtex_executable` is based on the `executable.yaml` and so can be changed just like any of the other executables if someone would prefer to use `biblatex` or `biber`.

source/lib/JMSLab/builders/build_latex.py
rcalvo12: @veli-m-andirin I believe this conversation can be resolved, let me know if there is anything else here that needs to happen.

source/lib/JMSLab/builders/build_latex.py
veli-m-andirin: Thanks @mcaceresb @rcalvo12!

@mcaceresb's list looks great to me. I am also fine with @rcalvo12's idea to exclude .blg. In case @mcaceresb agrees with excluding .blg, I think we can resolve this.

source/lib/JMSLab/builders/build_latex.py
veli-m-andirin: Thanks @rcalvo12, this implementation looks good to me. @mcaceresb, please feel free to resolve if it looks good to you, too.

veli-m-andirin:
Thanks @rcalvo12 for your persistent work here!

I think all my in-line comments are addressed (conditional on @mcaceresb's confirmation where I tagged him). I think it might be good to focus on the few broader points in https://github.com/JMSLab/Template/pull/56#issuecomment-1086312640 and https://github.com/JMSLab/Template/pull/56#pullrequestreview-929480958.

I also think it might be a good idea to take @jmshapir's and @mcaceresb's opinion about the following point in https://github.com/JMSLab/Template/pull/56#pullrequestreview-929480958:

> I wonder how we'd like to deal with cases where we have additional files that have references to the main text, such as an online Appendix. In that case, In these cases the order in which we compile the files would matter.

I am curious if we should have a builder that accounts for these cases. In theory, we can have cross-references across many files, and I am not sure about the architecture of a builder that can allow for an arbitrary number of auxiliary files such as online appendices. If we think that we should aim to account for the "most usual structure", then things might get simpler. If the most common structures are:

- A document with no appendices.
- A document with one appendix.

then a conditional logic to cover these would probably be simple. If we agree to have appendices in the same file as the main text as a lab practice, then we won't even need a conditional logic.

jmshapir:
Thanks @veli-m-andirin! Regarding the issue of multiple TeX files that you highlight in https://github.com/JMSLab/Template/pull/56#issuecomment-1099613997, my suggestion is to leave this aside for the purpose of this pull.

If no one has a current use case, we can perhaps add it to the roadmap as a [feature request](https://github.com/JMSLab/Template/wiki/Roadmap/17743f81f1d98feb04bc8c0dab1f06d5a1ea7cb9#features-to-consider).

If, now or later, either we in the lab or one of the [developers](https://github.com/JMSLab/Template/issues/53) has a use case, we can open an issue to tackle it. (I suspect there may be a way to handle arbitrary numbers of files but we might need to add a metadata file that tells the builder the order in which to compile. And we could always start with the simplest case as you say.)

Does that sound good?

veli-m-andirin:
Thanks @jmshapir! This sounds good to me. To be precise, the use case I had in mind was based on EventStudy where we had an appendix that cross-referenced the main text where the order of compilation mattered. That being said, I agree that having a builder that can compile a single file with a bibliography correctly would be a good first step.

source/lib/JMSLab/builders/build_latex.py
rcalvo12: Per call with @veli-m-andirin, in 8474693 I added the cleanup function before the compilation as well.

rcalvo12:
> If, now or later, either we in the lab or one of the [developers](https://github.com/JMSLab/Template/issues/53) has a use case, we can open an issue to tackle it. (I suspect there may be a way to handle arbitrary numbers of files but we might need to add a metadata file that tells the builder the order in which to compile. And we could always start with the simplest case as you say.)

> To be precise, the use case I had in mind was based on EventStudy where we had an appendix that cross-referenced the main text where the order of compilation mattered. That being said, I agree that having a builder that can compile a single file with a bibliography correctly would be a good first step.

@jmshapir I would be fine either way we do this. I can take a look at a simple appendix cross-reference case and see if I can come up with a solution given the EventStudy use case @veli-m-andirin pointed out (Maybe budget no more than a couple of hours to look at it). I am also fine leaving this for another issue down the road.

jmshapir:
Thanks @veli-m-andirin @rcalvo12!

My instinct is to leave the multi-TeX-file case for another issue.

mcaceresb:
@jmshapir Should tests be tackled in this PR or a later issue? If the latter, two notes:

- The LaTeX tests clash with the BibTeX addition, but this is probably fine to tackle later.
- There are no BibTeX specific tests. I think it's probably fine do this later as well so long as the builder has been tested non-formally.

LMK. My code comments are addressed but not the test comments I made [here](https://github.com/JMSLab/Template/pull/56#issuecomment-1086312640). However, if the tests are for a follow-up I'll approve as well.

jmshapir:
> @jmshapir Should tests be tackled in this PR or a later issue?

@mcaceresb thanks. My instinct is that, before we merge this pull, it's probably good to have at least some minimal working tests for the functionality we've implemented here, including TeX+BibTeX. Let me know if that answers or if we should make a more fine-grained decision.

@rcalvo12 fyi.

mcaceresb:
> > @jmshapir Should tests be tackled in this PR or a later issue?
> 
> @mcaceresb thanks. My instinct is that, before we merge this pull, it's probably good to have at least some minimal working tests for the functionality we've implemented here, including TeX+BibTeX. Let me know if that answers or if we should make a more fine-grained decision.
> 
> @rcalvo12 fyi.

@jmshapir Sounds good. 

@rcalvo12 If you know how to implement the above great; otherwise I am happy to go into more detail about how the tests here are structured.

rcalvo12:
> @rcalvo12 If you know how to implement the above great; otherwise I am happy to go into more detail about how the tests here are structured.

@mcaceresb I spent some time looking over it, and I think some more detail about how the tests are structured would be extremely useful. Thanks!

mcaceresb:
@rcalvo12 

## How the tests are structured

- The tests are based on the `unittest` module, which you cans see as we are basing the test classes on `unittest.TestCase`.

- In general to add a test we define a function named `def test_...` with the structure: 

    ```python
    @[patch]
    def test_[test](self, mock_[mock]):
        # test code
    ```

This much I reckon is clear from reading the code, but I think figuring out what is happening beyond this becomes difficult. Let me explain this by dissecting the parts of the first LaTeX test:

```
@subprocess_patch
def test_default(self, mock_check_output):
    '''
    Test that build_latex() behaves correctly when provided with
    standard inputs.
    '''
    mock_check_output.side_effect = fx.latex_side_effect
    target = 'build/latex.pdf'
    helpers.standard_test(self, build_latex, 'tex',
                          system_mock = mock_check_output,
                          source = ['test_script.tex'],
                          target = target)
    self.assertTrue(os.path.isfile(target))
```

- Why do all tests have a patch decorator and mock output? The first key thing to know is that we can only test the python code with the automated unit tests. In other words, we cannot in general run the entire repository (proprietary software such as Matlab and Stata is not available on-demand in gibhut servers; while we _could_ run LaTeX, LyX, and R tests, since they are open source, for now we are only testing the python code itself). 

    The solution is to "pretend" that we are running them without actually invoking the command-line call. The `@` decorator patches the system call so that it does not actually get invoked by the system; instead it gets invoked in a pretend testing environment. The `mock` function is to generate pretend output so we can check whether the python code is doing what it's supposed to (this gets generated in lieu of actual output).

- `@subprocess_patch` in this case patches the calls to `check_output` made by the builder and replaces them with mock calls.

- `mock_check_output` produces mock output based on a `side_effect`; these are program-specific and defined in `_side_effects.py`.

- The rest of the code is meant to check a particular aspect of the builder works as expected. In this case, `standard_test` is run, which is defined in `_test_helpers.py`. This calls the builder, redirects the log file, and runs the side-effect.

- Finally, we check what we want. In this case, we check the target file was created (this happens in the side-effect, which parses the function call using a pre-defined regex; see `commandd_match()` for details).

## High-level issues 

These should be resolved _before_ coding any new tests:

1. At the moment the standard tests call `system_mock.assert_called_once()` to check that the builder was only called once (since this is standard for all programs except LaTeX). We need to decide whether to make a LaTeX-specific exception or collapse the three LaTeX calls to a single `check_output` call.
    - My vote is to make a LaTeX-specific exception; we'd need to figure out how to have `system_mock` assert it was called exactly thrice for `.tex` files. Unsure how to do this but I assume it is possible. LMK if you do not find it in the docs.
    - While it's definitely possible to do a single `check_output` call for LaTeX with os-specific switches, the main drawback is that we'd have to modify the LaTeX regex, which I don't think is a good idea. This would be the case even if we found an OS-independent way to do it I think (unless `check_output` has a "repeat command" switch),

2. BibTeX cannot be checked with the current LeTeX tests for two reasons:
    - The LaTeX regex could not match the BibTeX command.
    - `check_output` would be called four times instead of three with BibTeX.

## Next steps

This is my own take:

1. Figure out how to make the LaTeX tests pass. This should require only adding the switch specific to `LaTeX` to assert three calls to `check_output`.

2. Decide how to code the BibTeX tests. There are two choices:
    - Have a BibTeX-specific file.
    - Add it to the LaTeX tests.

    My vote is for the latter. In either case we need to code a BibTeX side-effect. For this step, it will be sufficient to code an empty BibTeX side-effeect and add logic in the LaTeX side-effect to switch to BibTeX when this is called instead of LaTeX. LMK if you have trouble doing this (it's fine for the side-effect to be empty since none of the LaTeX tests should call it).

Once that is done, if both are correct, we can:

3. Add a single BibTeX test, whcih just checks the function calls are as expected. This is the hardest step.
    - You have to now code a BibTeX side effect that creates a `.blg` and `.bbl` file based on the BibTeX command.
    - The BibTeX command has to be parsed by a BibTeX-specific regex, which you will have to code an add to `commandd_match`.
    - There should be another switch so `check_output` is asserted to have been called 4 times (not 3).
    - I can also explain the side-effect in detail when you get to this step.

I think this will be plenty for this issue. You can add a `test_bibtex_basic` function to the LaTeX tests and add a `.bib` file to the dependenties, which should activate the `BibTeX`. This tests:

- That the builder without `.bib` is correct (the rest of the LaTeX tests would do this).
- The builder with `.bib` is correct (this new BibTeX tests would do this).

The other two tests to add would be:

4. Test multiple `.bib` files.
5. Test the output having a different name than the input.

But I think 1-3 above is already plenty for this PR.

rcalvo12:
@mcaceresb Thank you for the incredibly useful summary above. I tried my hand at implementing the first step so that it asserts 3 calls if latex is specified. Any feedback would be appreciated.

mcaceresb:
> @mcaceresb Thank you for the incredibly useful summary above. I tried my hand at implementing the first step so that it asserts 3 calls if latex is specified. Any feedback would be appreciated.

@rcalvo12 Nice! The only issue that remained was that LaTeX cleanup is invoked on failure and expects `add_out_name` to have been run, so I switched it to L32 (above `execute_system_call`). You can see all the tests are now passing. This should always be the case now (you can run tests locally using `pytest`).

Happy to continue to help as you make progress on coding the BibTeX test.

rcalvo12:
@mcaceresb Thanks again for your help here! I'm trying to work on this step now:

> In either case we need to code a BibTeX side-effect. For this step, it will be sufficient to code an empty BibTeX side-effect and add logic in the LaTeX side-effect to switch to BibTeX when this is called instead of LaTeX. LMK if you have trouble doing this (it's fine for the side-effect to be empty since none of the LaTeX tests should call it).

I was able to add an empty side-effect but I am unclear on how to add the logic within the LaTeX side effect for switching. Maybe you can expand on this a bit more?

rcalvo12:
@mcaceresb I tried my hand at what you suggested [here](https://github.com/JMSLab/Template/pull/56#issuecomment-1112455143), would appreciate feedback. Thanks!

mcaceresb:
@rcalvo12 One excellent resource is [regex101](https://regex101.com). You can input your regex and sample text to see if it matches. If you select `python` as the flavor, you can see

```
(?P<executable>\w+)\s+(?P<target>-\w+\s+\S+)?\s*(?P<log_redirect>\>\s*[\.\/\\\w]+\.\w+)?
```

does not quite partition

```
bibtex target_file > sconscript.log
```

correctly. While the expression matches, the groups are not divided in the way we want. You can see the issue is with `target`, which was based on a regex for an option rather than a source or target. (Note other options for sources/targets assume a file extension.) You can play around with the website, or see what worked for me:

<details>
  <summary>spoiler</summary>

  ```
  (?P<executable>\w+)\s+(?P<target>[\.\/\\\w]+)?\s*(?P<log_redirect>\>\s*[\.\/\\\w]+\.\w+)?
  ```

(This accounts for slashes in the file name as well. Note you'd have to put this back into the format in `_test_helpers.py`.)

</details>

PS: I do realize I'm giving you feedback in a very piece-meal way. I thought it would be helpful but LMK in case you want me to more actively help in finishing to code this: I don't want to prompt you to spend too much time on it if it's not helpful for you.

rcalvo12:
@mcaceresb After spending some time on this issue over the past handful of weeks, I am not much closer on the implementation of the bibtex test and I think it would probably be more efficient to bring you in on the implementation. Maybe we can set up a call to discuss, so I can learn from you more directly on this.

mcaceresb:
@rcalvo12 Messaged you; we'll set something up and hopefully close this PR.

@jmshapir fyi

mcaceresb (approved):

mcaceresb:
@rcalvo12 Following our meeting, I pushed the changes we discussed implementing:

- The mock bibtex test.
- The `nsyscalls` argument to `standard_test` to account for LaTeX and `bibtex` running the builder multiple times.
- An assertion that the `check_bib` function works when `.bib` files are specified.

The automated python tests are all passing and I've approved the PR.

@jmshapir fyi

rcalvo12:
Thanks @mcaceresb for that very productive meeting! In https://github.com/JMSLab/Template/pull/56/commits/58010a8070f96f59c224c82cb75b2f57f39fc443, I reran the code and tests on my end and I also added latex to the list of envs in `Sconstruct`. Everything appears to be working as intended.

@jmshapir fyi.

rcalvo12:
Removed issue sub in https://github.com/JMSLab/Template/pull/56/commits/537fcfd63c7dd91f7b0ec4345c3c3bae428c2984.

output/analysis/plots/sconscript.log
```diff
@@ -1,8 +1,74 @@
-*** Builder log created: {2022-05-12 12:55:01}
-*** Builder log completed: {2022-05-12 12:55:24}
-[Warning: Function plot has the same name as a MATLAB builtin. We suggest you rename the function to avoid a potential name conflict.] 
-[> In path (line 109)
-In addpath (line 86)] 
+*** Builder log created: {2022-08-24 12:24:43}
+*** Builder log completed: {2022-08-24 12:25:07}
+[Warning: The following error was caught while executing 'onCleanup' class
+destructor:
```
jmshapir, line 4: Note errors here. Might just be warnings due to system differences but wanted to flag.

output/derived/wb_clean/sconscript.log
```diff
@@ -1,54 +1,2 @@
...
-. main
-(3 vars, 264 obs)
-(12 missing values generated)
-(100 missing values generated)
-file output/derived/wb_clean/gdp_education_logs.csv saved
-
-. 
-end of do-file
+*** Builder log created: {2022-08-24 12:23:23}
+*** Builder log completed: {2022-08-24 12:23:32}
```
jmshapir, line 2: Intentional that Stata log is not included here?

output/paper/TemplateTex.blg
jmshapir, line 1: Maybe gitignore the BLG file?

Also reminder to ignore the PDF files since we can't use git-lfs tracking in a Template.

output/paper/TemplateTex.log
```diff
@@ -0,0 +1,394 @@
+This is pdfTeX, Version 3.141592653-2.6-1.40.22 (TeX Live 2021) (preloaded format=pdflatex 2022.2.4)  24 AUG 2022 12:25
```
jmshapir, line 1: Maybe gitignore this LOG file? I think the info anyway ends up in the `sconscript.log`.

jmshapir (changes requested):
Thanks @rcalvo12 @mcaceresb!

I updated the branch from `main` and made a few comments.

Apologies if we've already discussed some of these points about what to ignore/keep.

Thanks!

output/paper/TemplateTex.log
rcalvo12: This was already in the gitignore but had to be removed from the repo. Done in https://github.com/JMSLab/Template/pull/56/commits/1aafefda0f0c9a82a09ae0cffa06b458611081d6.

output/paper/TemplateTex.blg
rcalvo12: The blg file was already in the gitignore but had to be removed from the repo. Done in https://github.com/JMSLab/Template/pull/56/commits/1aafefda0f0c9a82a09ae0cffa06b458611081d6. I also removed the pdfs in https://github.com/JMSLab/Template/pull/56/commits/1aafefda0f0c9a82a09ae0cffa06b458611081d6 and added the pdfs in output/paper to the gitignore in https://github.com/JMSLab/Template/pull/56/commits/6b0f747a00d6a695a35108a22be5fa18ec354842.

rcalvo12: Actually, I changed how pdfs are ignored in https://github.com/JMSLab/Template/pull/56/commits/9af7f8d289b1be1c3ea5bbdb698cf368deb73936, so that rather than ignoring all .pdfs in output/paper, I just ignore the specific template pdfs by name. I also removed the other remaining pdfs in the repo in https://github.com/JMSLab/Template/pull/56/commits/9af7f8d289b1be1c3ea5bbdb698cf368deb73936.

@jmshapir fyi.

output/derived/wb_clean/sconscript.log
rcalvo12: This is just because the output files were already built when scons went to build it. When this happens and no changes were made, scons just moves on rather than rebuilding it. I removed the outputs in https://github.com/JMSLab/Template/pull/56/commits/9af7f8d289b1be1c3ea5bbdb698cf368deb73936 and reran the data and the log was restored.

output/analysis/plots/sconscript.log
rcalvo12: Noted. When I removed the outputs and reran in https://github.com/JMSLab/Template/pull/56/commits/9af7f8d289b1be1c3ea5bbdb698cf368deb73936, the errors went away. Not sure where they originated, have not been able to reproduce. My guess is it was some sort of system difference as you noted.

rcalvo12:
@veli-m-andirin Did you have any other changes here? If not can you approve?

source/lib/JMSLab/tests/output_macros.tex
```diff
@@ -0,0 +1,2 @@
+\newcommand{\Epsilon}{-1.19}
```
veli-m-andirin, line 1: @rcalvo12, is this test used somewhere?

veli-m-andirin (commented):
Thanks @rcalvo12 @mcaceresb for great work here!

@rcalvo12, I left one in-line comment. Also, do you know why the talk is removed from output/talk?

veli-m-andirin:
@rcalvo12, feel free to ignore the comment about the pdf for talk, it seems to be done on purpose!

jmshapir:
> @rcalvo12, feel free to ignore the comment about the pdf for talk, it seems to be done on purpose!

Thanks @veli-m-andirin! See https://github.com/JMSLab/Template/issues/66#issue-1378677265 for context.

source/lib/JMSLab/tests/output_macros.tex
rcalvo12: @veli-m-andirin From looking into it, it seems like this file was a temp test file produced by `test_autofill.py` but is supposed to be cleaned up. My best guess is that somewhere along the way as we were working on testing and this script was run but was interrupted and so the file was produced but not cleaned up. I removed the file in https://github.com/JMSLab/Template/pull/56/commits/3710921502e0e46307ca1104fe5653a92ef3effe and had no problem re-running scons or the testing. Thanks for the catch!

veli-m-andirin (approved):
Thanks @rcalvo12! Approving.

rcalvo12:
Summary in https://github.com/JMSLab/Template/issues/42#issuecomment-1256343749.

Issue #55, opened 2022-03-25
Add and refine unit tests for autofill functions


jinfu93:
Porting over from the [sketch](https://github.com/JMSLab/Template/pull/51#discussion_r827376226) in #51 below are the additional tests we would like to implement in `test_autofill.py`:

- [x] Test that the function creates a new file.
- [x] Test that the file created by the function contains the correct command names and values.
- [x] What is the expected behavior of the function when a variable name is not found?
  - Add error handling.
  - Construct a test to verify error handling works as expected.
- [x] Make sure the behavior is as expected when using a string for format and a list with different formats.
  - Should we allow for a default format, so that the function knows what to do when `autofill_formats` is not specified?
- [x] Streamline test suite

@santiagohermo do feel free to add to this list or expand on the details of each test, thanks a lot!

santiagohermo:
> @santiagohermo do feel free to add to this list or expand on the details of each test, thanks a lot!

Thanks @jinfu93! I made a couple of edits to the [opening comment](https://github.com/JMSLab/Template/issues/55#issue-1181188969).

jinfu93:
@santiagohermo I've added drafts of 2 tests:
https://github.com/JMSLab/Template/blob/442eb18e64e6348e8768987c238b150b9140c874/source/lib/JMSLab/tests/test_autofill.py#L26-L44

Let me know if these look different from what you had in mind!

Regarding:
>Make sure the behavior is as expected when using a string for format and a list with different formats.

I'm thinking of adding error handling to accommodate when `autofill_formats` doesn't line up with `autofill_lists` i.e. if `autofill_lists` is a nested list but `autofill_formats` is a string. Subsequently, the test can ensure the exception is thrown as expected. Let me know if you were thinking of something more straightforward than that.

I've also added a default for `autofill_formats` in [1a52e3e](https://github.com/JMSLab/Template/commit/1a52e3e3b47d2031c2a7ad728a3d6144ff9adb4e) as you suggested.

Thanks!

jinfu93:
@santiagohermo implemented the following:
>I'm thinking of adding error handling to accommodate when autofill_formats doesn't line up with autofill_lists i.e. if autofill_lists is a nested list but autofill_formats is a string. Subsequently, the test can ensure the exception is thrown as expected. Let me know if you were thinking of something more straightforward than that.

[...]

PR #57, opened 2022-04-10
Pull Request for #55: Add and refine unit tests for autofill functions

jinfu93:
@santiagohermo significant changes here are contained in `source/lib/JMSLab/tests/test_autofill.py` where we added 3 additional tests on top of checking for out file existence (as per this [list](https://github.com/JMSLab/Template/issues/55#issue-1181188969)).

Also added default format for autofill wrapper in `source/lib/JMSLab/autofill.py`.

Thanks!

### DIFF
--- a/source/lib/JMSLab/autofill.py
+++ b/source/lib/JMSLab/autofill.py
@@ -19,7 +19,7 @@ def Autofill(var, format = "{}", namespace = None):
     return newcommand.format(commandname, content)
 
 def GenerateAutofillMacros(autofill_lists, autofill_formats = "{:.2f}", autofill_outfile = "autofill.tex"):
-    """
+    '''
 
     Parameters
     ----------
@@ -31,7 +31,7 @@ def GenerateAutofillMacros(autofill_lists, autofill_formats = "{:.2f}", autofill
     -------
     .tex file
 
-    """
+    '''
     if type(autofill_lists) != list:
             raise Exception("Argument 'autofill_lists' must be list")
             
--- a/source/lib/JMSLab/autofill.py
+++ b/source/lib/JMSLab/autofill.py
@@ -10,7 +10,7 @@ def Autofill(var, format = "{}", namespace = None):
             parent = parent.f_back
         
         if var not in parent.f_locals.keys() or parent is None:
-            raise Warning(f"Autofill: Variable '{var}' not found")
+            raise Exception(f"Autofill: Variable '{var}' not found")
         
         namespace = parent.f_locals
     
--- a/source/lib/JMSLab/autofill.py
+++ b/source/lib/JMSLab/autofill.py
@@ -37,8 +37,8 @@ def GenerateAutofillMacros(autofill_lists, autofill_formats = "{:.2f}", autofill
             
     nested_list = any(isinstance(i, list) for i in autofill_lists)
 
-    if (nested_list == True and type(autofill_formats) == str) or (nested_list == False 
-                                                                    and type(autofill_formats) == list):
+    if ((nested_list and type(autofill_formats) is str) 
+        or (nested_list and type(autofill_formats) is list)):
         raise Exception("Arguments 'autofill_lists' and 'autofill_formats' are incompatible")
 
     autofill_file = open(autofill_outfile, 'w')
--- a/source/lib/JMSLab/autofill.py
+++ b/source/lib/JMSLab/autofill.py
@@ -2,7 +2,7 @@ import inspect
 
 def Autofill(var, format = "{}", namespace = None):
     newcommand = f"\\newcommand{{{{\\{{}}}}}}{{{{{format}}}}}\n"
-    
+
     # Look for var in parent frames
     if namespace is None:
         parent = inspect.currentframe().f_back
@@ -13,14 +13,14 @@ def Autofill(var, format = "{}", namespace = None):
             raise Exception(f"Autofill: Variable '{var}' not found")
         
         namespace = parent.f_locals
-    
+
     commandname, content = var, namespace[var]
-    
+
     return newcommand.format(commandname, content)
 
 def GenerateAutofillMacros(autofill_lists,  autofill_outfile, autofill_formats = "{:.2f}"):
     """
-    
+
     Parameters
     ----------
     autofill_lists : TYPE - list
@@ -36,13 +36,13 @@ def GenerateAutofillMacros(autofill_lists,  autofill_outfile, autofill_formats =
             raise Exception("Argument 'autofill_lists' must be list")
             
     nested_list = any(isinstance(i, list) for i in autofill_lists)
-    
+
     if (nested_list == True and type(autofill_formats) == str) or (nested_list == False 
-                                                                   and type(autofill_formats) == list):
+                                                                    and type(autofill_formats) == list):
         raise Exception("Arguments 'autofill_lists' and 'autofill_formats' are incompatible")
-            
+
     autofill_file = open(autofill_outfile, 'w')
-    
+
     if type(autofill_formats) == str:
         output_macros = ''.join(Autofill(autofill_var, autofill_formats) for autofill_var in autofill_lists)
     else:
--- a/source/lib/JMSLab/autofill.py
+++ b/source/lib/JMSLab/autofill.py
@@ -38,7 +38,7 @@ def GenerateAutofillMacros(autofill_lists, autofill_formats = "{:.2f}", autofill
     nested_list = any(isinstance(i, list) for i in autofill_lists)
 
     if ((nested_list and type(autofill_formats) is str) 
-        or (nested_list and type(autofill_formats) is list)):
+        or (not nested_list and type(autofill_formats) is list)):
         raise Exception("Arguments 'autofill_lists' and 'autofill_formats' are incompatible")
 
     autofill_file = open(autofill_outfile, 'w')
--- a/source/lib/JMSLab/autofill.py
+++ b/source/lib/JMSLab/autofill.py
@@ -18,7 +18,7 @@ def Autofill(var, format = "{}", namespace = None):
 
     return newcommand.format(commandname, content)
 
-def GenerateAutofillMacros(autofill_lists,  autofill_outfile, autofill_formats = "{:.2f}"):
+def GenerateAutofillMacros(autofill_lists, autofill_formats = "{:.2f}", autofill_outfile = "autofill.tex"):
     """
 
     Parameters
--- a/source/lib/JMSLab/autofill.py
+++ b/source/lib/JMSLab/autofill.py
@@ -18,7 +18,7 @@ def Autofill(var, format = "{}", namespace = None):
     
     return newcommand.format(commandname, content)
 
-def GenerateAutofillMacros(autofill_lists,  autofill_formats, autofill_outfile):
+def GenerateAutofillMacros(autofill_lists,  autofill_outfile, autofill_formats = "{:.2f}"):
     """
     
     Parameters
--- a/source/lib/JMSLab/autofill.py
+++ b/source/lib/JMSLab/autofill.py
@@ -32,6 +32,15 @@ def GenerateAutofillMacros(autofill_lists,  autofill_outfile, autofill_formats =
     .tex file
 
     """
+    if type(autofill_lists) != list:
+            raise Exception("Argument 'autofill_lists' must be list")
+            
+    nested_list = any(isinstance(i, list) for i in autofill_lists)
+    
+    if (nested_list == True and type(autofill_formats) == str) or (nested_list == False 
+                                                                   and type(autofill_formats) == list):
+        raise Exception("Arguments 'autofill_lists' and 'autofill_formats' are incompatible")
+            
     autofill_file = open(autofill_outfile, 'w')
     
     if type(autofill_formats) == str:
--- a/source/lib/JMSLab/tests/test_autofill.py
+++ b/source/lib/JMSLab/tests/test_autofill.py
@@ -5,54 +5,54 @@ from os.path import exists
 from ..autofill import GenerateAutofillMacros
 
 class Test(TestCase):
-    
+
     def setUp(self):
         self.tempdir = tempfile.mkdtemp() 
         self.outfile = self.tempdir + r"output_macros.tex"
         return
-    
+
     def tearDown(self):
         shutil.rmtree(self.tempdir)
         return
-    
+
     def test_file_exists(self):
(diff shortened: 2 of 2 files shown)


jmshapir:
@santiagohermo thanks in advance for your review of this pull!

santiagohermo:
Thanks @santiagohermo @jinfu93! Task received, I'll get to work on it probably tomorrow. (I should have flagged my plans before, apologies!)

source/lib/JMSLab/autofill.py
```diff
@@ -10,15 +10,15 @@ def Autofill(var, format = "{}", namespace = None):
...
+            raise Exception(f"Autofill: Variable '{var}' not found")
         
         namespace = parent.f_locals
     
     commandname, content = var, namespace[var]
     
     return newcommand.format(commandname, content)
 
-def GenerateAutofillMacros(autofill_lists,  autofill_formats, autofill_outfile):
+def GenerateAutofillMacros(autofill_lists,  autofill_outfile, autofill_formats = "{:.2f}"):
```
santiagohermo, line 21: The order of the arguments looks weird here, but not sure how to fix it!

Would there be a natural default for `autofill_outfile` so that we can put it in the end? Why not `autofill_outfile = '.'`?

source/lib/JMSLab/autofill.py
```diff
@@ -10,15 +10,15 @@ def Autofill(var, format = "{}", namespace = None):
...
     """
```
santiagohermo, line 22: Minor: could we use `'''` for consistency in style with other scripts? Eg, see [log.py](https://github.com/JMSLab/Template/blob/db8ac17cbb45a65ec967d7bfdf5af94afc0c6329/source/lib/JMSLab/log.py)

source/lib/JMSLab/autofill.py
```diff
@@ -32,6 +32,15 @@ def GenerateAutofillMacros(autofill_lists,  autofill_formats, autofill_outfile):
     .tex file
 
     """
+    if type(autofill_lists) != list:
+            raise Exception("Argument 'autofill_lists' must be list")
```
santiagohermo, line 36: Just asking: does this stop the execution of the program?

source/lib/JMSLab/autofill.py
```diff
@@ -32,6 +32,15 @@ def GenerateAutofillMacros(autofill_lists,  autofill_formats, autofill_outfile):
...
+            
+    nested_list = any(isinstance(i, list) for i in autofill_lists)
+    
+    if (nested_list == True and type(autofill_formats) == str) or (nested_list == False 
+                                                                   and type(autofill_formats) == list):
```
santiagohermo, line 41: You can replace `nested_list == True` with just `nested_list` and `nested_list == False` with `not nested_list`. You can also replace the `==` in `type(autofill_formats) == str` and `type(autofill_formats) == list` with `is` and you end up with a sentence you can almost read :)

Also, it may look a bit nicer to break the line on the `or`?

source/lib/JMSLab/tests/test_autofill.py
```diff
@@ -1,25 +1,64 @@
...
-
-sys.path.append('./source/lib')
-
 from unittest import main, TestCase
-from JMSLab.autofill import GenerateAutofillMacros
+import tempfile, shutil 
 from os.path import exists
 
-
+from ..autofill import GenerateAutofillMacros
```
santiagohermo, line 6: Style-related: should we have all the `import stuff` together and the `from a import B` together, as in for instance [test_build_latex.py](https://github.com/JMSLab/Template/blob/db8ac17cbb45a65ec967d7bfdf5af94afc0c6329/source/lib/JMSLab/tests/test_build_latex.py)?

source/lib/JMSLab/tests/test_autofill.py
```diff
@@ -1,25 +1,64 @@
...
 
 class Test(TestCase):
+    
+    def setUp(self):
+        self.tempdir = tempfile.mkdtemp() 
+        self.outfile = self.tempdir + r"output_macros.tex"
+        return
```
santiagohermo, line 12: I don't understand this return statement. Did you mean this?

```suggestion
        return None
```

There are similar 'lonely' return statements below.

source/lib/JMSLab/autofill.py
```diff
@@ -32,6 +32,15 @@ def GenerateAutofillMacros(autofill_lists,  autofill_formats, autofill_outfile):
     .tex file
 
     """
+    if type(autofill_lists) != list:
+            raise Exception("Argument 'autofill_lists' must be list")
+            
+    nested_list = any(isinstance(i, list) for i in autofill_lists)
```
santiagohermo, line 38: I'm not sure in what context we may want `autofill_lists` to be a nested list?

santiagohermo (changes requested):
Great work @jinfu93! Those tests pass correctly on my computer. I left a few comments and suggestions below.

I'm getting two warnings when running the tests though, which I thought might be helpful to flag to @jmshapir.

image

source/lib/JMSLab/tests/test_autofill.py
santiagohermo: Trying out the code suggestions here @jmshapir. So cool!

source/lib/JMSLab/autofill.py
jinfu93: Thanks @santiagohermo! Took a pass at it in [af4bbde](https://github.com/JMSLab/Template/pull/57/commits/af4bbde924d7676b5b3553f1dbc38dfbcc387c27) let me know if it makes sense.

source/lib/JMSLab/autofill.py
jinfu93: @santiagohermo do you mean to replace all the "" with ''? I still see some discrepancies in log.py for instance:
https://github.com/JMSLab/Template/blob/db8ac17cbb45a65ec967d7bfdf5af94afc0c6329/source/lib/JMSLab/log.py#L17

vs

https://github.com/JMSLab/Template/blob/db8ac17cbb45a65ec967d7bfdf5af94afc0c6329/source/lib/JMSLab/log.py#L23

Do let me know if I misunderstood, thanks!

source/lib/JMSLab/autofill.py
jinfu93: @santiagohermo yes indeed it does! Do we want to raise a warning instead?

source/lib/JMSLab/autofill.py
jinfu93: @santiagohermo sorry that it's not very clear from the get go, but the original use case that this function emerged from was something along the lines of this:
`
GenerateAutofillMacros([["Epsilon", "MarginalCost"],["CartelProfits", "CartelCosts"]], 
["{:.2f}", "\\textnormal{{{:.2f}}}"], 
"output_macros.tex")
`

where we have 2 sets of floats that we want to autofill in different formats. Let me know if that makes sense, thanks!

source/lib/JMSLab/tests/test_autofill.py
jinfu93: Addressed remaining lonely return statements in [e6fc892](https://github.com/JMSLab/Template/pull/57/commits/e6fc89248dfdc62f74fbb214b74f33bd346e4dc8).

source/lib/JMSLab/autofill.py
jinfu93: Thanks @santiagohermo for the pointers! Addressed in [8915073](https://github.com/JMSLab/Template/pull/57/commits/8915073e00ed687c39568d2f2f5a058c9eed1a26).

source/lib/JMSLab/autofill.py
santiagohermo: Thanks @jinfu93! I see that we are not consistent with using `'` or `"` throughout. In [the comment](https://github.com/JMSLab/Template/pull/57#discussion_r851325117) I just meant to use `'''` instead of `"""` to open and close the comment that explained the function, ie, [this one](https://github.com/JMSLab/Template/blob/db8ac17cbb45a65ec967d7bfdf5af94afc0c6329/source/lib/JMSLab/autofill.py#L22-L34)

source/lib/JMSLab/autofill.py
santiagohermo: Look good, thanks @jinfu93!

source/lib/JMSLab/autofill.py
santiagohermo: Oh, got it. Thanks @jinfu93, that makes sense.

Would we want to test this case in [`test_output`](https://github.com/JMSLab/Template/blob/d85dbd8546607d213eea620763faa2792e518219/source/lib/JMSLab/tests/test_autofill.py#L33)?

source/lib/JMSLab/autofill.py
jinfu93: Got it thanks for clarifying! Change made in [6ab726e](https://github.com/JMSLab/Template/pull/57/commits/6ab726e3e1f3fd710092d68c81ff73b9e6d2019d).

source/lib/JMSLab/autofill.py
jinfu93: Great idea @santiagohermo! Added in [0eb794e](https://github.com/JMSLab/Template/pull/57/commits/0eb794ee6f892d4e2fdb6f3a6fad127f024f5624).

jinfu93:
Thanks @santiagohermo for iterating on the details carefully with me! Appreciate all the very helpful comments and tips. I think we might be done with the remaining open points on your review but do let me know if that's not the case. 

Should we try to tackle the warnings as per your [above comment](https://github.com/JMSLab/Template/pull/57#pullrequestreview-943489048) as well? I was wondering if you have any ideas what might be causing them. Thanks!

source/lib/JMSLab/autofill.py
santiagohermo: Thanks @jinfu93! Just noting that we could both cases here.

santiagohermo (approved):
> Thanks @santiagohermo for iterating on the details carefully with me! Appreciate all the very helpful comments and tips. I think we might be done with the remaining open points on your review but do let me know if that's not the case. 
> 
> Should we try to tackle the warnings as per your [above comment](https://github.com/JMSLab/Template/pull/57#pullrequestreview-943489048) as well? I was wondering if you have any ideas what might be causing them. Thanks!

Thanks @jinfu93 for the detailed explanation! All my comments were addressed (except a minor thing [here](https://github.com/JMSLab/Template/pull/57#discussion_r853234791) which you can resolve as you think more convenient), and after updating the branch the tests passed. Thus, I'm approving.

I still get the warnings. They may be machine-specific though. Do you get them as well? Happy to try solving them here.

santiagohermo:
On the tests, from the screenshot [here](https://github.com/JMSLab/Template/pull/57#pullrequestreview-943489048). 

- It looks like the first one comes from a package so not sure we can do a lot there. 
- The second seems to be caused by [this function](https://github.com/JMSLab/Template/blob/fff134e5d649825bf81801c2a80e2acb8f8d2c0c/source/lib/JMSLab/tests/test_log.py#L130-L150), and in particular I would guess [L144](https://github.com/JMSLab/Template/blob/fff134e5d649825bf81801c2a80e2acb8f8d2c0c/source/lib/JMSLab/tests/test_log.py#L144). I conjecture that changing the `1.0` to `1` in that line will remove the warning. However, we should make sure the `1.0` is not essential there.

jinfu93:
Thanks for investigating @santiagohermo! I do get the same warnings as well. I tried your suggested fix but it seems like `1.0` was essential for the test.

image

Should we dive more into `test_log.py` to investigate?

santiagohermo:
Thanks @jinfu93! I made a little bit of digging on both warnings (flagged in https://github.com/JMSLab/Template/pull/57#pullrequestreview-943489048).

The first warning has been flagged in https://github.com/pyreadline/pyreadline/issues/65. I expect this to solve itself for future versions of this package.

The second warning arises in line 144 of the below

https://github.com/JMSLab/Template/blob/fff134e5d649825bf81801c2a80e2acb8f8d2c0c/source/lib/JMSLab/tests/test_log.py#L130-L150

The reason that changing `1.0` for `1` doesn't work is related to the note in the function's description. I would recommend to just drop lines 143-144. An alternative would be to add some error handling in `log.py` for the cases when the argument `log` of `start_log()` is an integer or a boolean, but this may be overkill. What do you think? We can also ask Mauricio his opinion here.

jinfu93:
Thanks for digging into this @santiagohermo!

> The first warning has been flagged in https://github.com/pyreadline/pyreadline/issues/65. I expect this to solve itself for future versions of this package.

This sounds good to me as well.

> The reason that changing 1.0 for 1 doesn't work is related to the note in the function's description.

Great catch! I should've read the description myself earlier.

@mcaceresb would you happen to be familiar with the 2nd warning [here](https://github.com/JMSLab/Template/pull/57#pullrequestreview-943489048)? @santiagohermo made some suggestions [above](https://github.com/JMSLab/Template/pull/57#issuecomment-1103298953) and we were wondering if you might have an idea of how best to proceed? Thanks a lot!

mcaceresb:
@jinfu93 @santiagohermo The warnings should be fine. The first is a deprecation warning for a package that should be OK to use so long as it's upgraded whenever python is upgraded (i.e. if you go to Python 3.9+ you should also upgrade all packages).

The second refers to the fact you can pass an integer instead of a file name to `open`. One if the tests passes 1.0 or similar and expects it to fail since that's not a valid input. Python is saying that it can try to convert the float to an integer so that `open` doesn't fail but this behavior is deprecated.  I included a note about it in the [test description](https://github.com/JMSLab/Template/blob/main/source/lib/JMSLab/tests/test_log.py#L136-L140).

I don't think you can get rid of the first warning. I think you can get rid of the second by passing e.g. `1.1` instead of `1.0` to the log in the log test, but it shouldn't be necessary (since the test is supposed to fail).

jinfu93:
Thanks so much for your help looking into this @mcaceresb, that makes a lot of sense! I'm good to leave as is if you are @santiagohermo, in which case I think we can wrap up here? 

@jmshapir feel free to let us know if you have any further thoughts before we conclude here.

Thanks everyone!

santiagohermo:
Thanks for the comments @mcaceresb! I agree @jinfu93, let's wrap up here.

jinfu93:
Summary for associated issue can be viewed [here](https://github.com/JMSLab/Template/issues/55#issuecomment-1104353740).

Issue #69, opened 2022-10-26
Stray rscript.log


jmshapir:
@rcalvo12 @ew487 am I right that the file [here](https://github.com/JMSLab/Template/blob/8236cfde2884e945cb5758f6b8b40264bcf8a4c7/rscript.log) is an accidental commit?

If so I'll go ahead and remove it, thanks!

rcalvo12:
Thanks @jmshapir, Yes, this looks like an accidental commit! I don't think we ever use rscript.log files.

jmshapir:
Summary: In this issue we deleted a [stray file](https://github.com/JMSLab/Template/issues/69#issue-1424035188). Changes brought to `main` in [26a8a8e](https://github.com/JMSLab/Template/commit/26a8a8e6bc649553ff66d2a65eda30adca1e8111).

PR #70, opened 2022-10-26
Delete stray file for #69

jmshapir:
@rcalvo12 this branch

image

looks to be associated with an inactive issue.

Can I delete?


rcalvo12 (approved):
@jmshapir I don't think there should be any problems with removing this file so I am approving!

jmshapir:
Thanks @rcalvo12!

What do you think about the legacy branch I flagged in https://github.com/JMSLab/Template/pull/70#issue-1424152867?

rcalvo12:
@jmshapir, I don't believe there should be any problem with deleting this issue branch. It seems to have been created but have no commits made to it after being created. To be safe here is a permalink to its current state, incase we ever need to revisit: [66-test](https://github.com/JMSLab/Template/tree/441c75a1bbaeba035ef91d8eb145b2b95218ee36).

jmshapir:
Thanks @rcalvo12! I am guessing I was using that stray branch to test git ignore or git lfs functionality locally.

In any case I have deleted, thanks for preserving the stable link!

PR #73, opened 2023-01-04
PR for #72: Updated DAG to fix gap preventing `-j` parallelization from working

### LINKED ISSUE #72: Test parallelizing scons run
`scons` represents the build steps as a [directed acyclic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph).

This means it can calculate which paths can be built in parallel and which must be built in sequence.

The [`-j` option](https://scons.org/doc/production/HTML/scons-man.html#:~:text=scons%20supports%20building,parallel%2C%20for%20example.) is designed to allow a user to run multiple simultaneous tasks.

We'd like to explore whether it's possible to use this option successfully on FASRC.

MosesStewart:
Closes #72

Created [`source/figures/SConscript`](https://github.com/JMSLab/Template/blob/f6297fe96d4feb3d43f82c38edc2c96ef73add46/source/figures/SConscript) which contains code to let `scons` know that `#source/figures/gdp_educ.lyx` and `#source/figures/gdp_educ.tex` are dependent on `#output/analysis/plots/gdp_educ.eps`.  Also updated `SConstruct` to contain a call to `source/figures/SConscript`

This prevents `scons` from building `gdp_educ.eps` and `template.pdf` in parallel when using the `-j` option, opening up parallelization in the repository.

Adding @jmshapir and @rcalvo12 as reviewers per https://github.com/JMSLab/Template/issues/72#issuecomment-1371515874


source/figures/SConscript
```diff
@@ -0,0 +1,11 @@
+Import('*')
+
+target = ['#source/figures/gdp_educ.lyx']
+source = ['#output/analysis/plots/gdp_educ.eps']
+
+env.Depends(target = target, dependency = source)
```
jmshapir, line 6: I did not know about this trick for specifying a dependency without specifying a build operation. Neat!

jmshapir (approved):
Nice work here @MosesStewart!

rcalvo12 (approved):
@MosesStewart really awesome sleuthing here. I agree with @jmshapir that the way you built the `source/figures/Sconscript` is a very cool find! This definitely gives some important clues for how we would parallelize the scons builds in other, more complex repos. 

All the changes here look good to me so I am approving!

MosesStewart:
summary in https://github.com/JMSLab/Template/issues/72#issuecomment-1372682099

PR #86, opened 2023-11-08
PR for #85: Explore whether we can change file path structure for latex files

### LINKED ISSUE #85: Explore whether we can change file path structure for LaTeX files
In LyX we use paths relative to the LyX file:
https://github.com/JMSLab/Template/blob/380e12f4d8cb94165fd9f86262894780b008be1e/source/paper/Template.lyx#L272

In LaTeX we use paths relative to the root:
https://github.com/JMSLab/Template/blob/380e12f4d8cb94165fd9f86262894780b008be1e/source/paper/Template.tex#L53

The result is that, in a repository configured according to the `Template`, it's harder to compile manually in LaTeX than it is in LyX.

Can we investigate whether there is a robust way to modify the builder so that we can use the same path structure in LaTeX that we use in LyX?

Thanks!

ew487:
In this issue we moved the folder paths to the latex preamble, for latex files.
Closes #85


ew487:
@jmshapir There seems to be an issue with the references in the scons run, which you can see [here](https://draftable.com/compare/rZsCZVgudWLv) so I could take a look at that. I could also look at the problems with the failed checks?

source/paper/Template.tex
```diff
@@ -41,15 +46,15 @@ \section{Some Section}
...
 \newpage{}
 
-\input{source/figures/gdp_educ}
+\input{gdp_educ}
 
 \newpage{}
 \appendix
 \bibliographystyle{apalike}
-\bibliography{source/paper/References}
+\bibliography{References}
```
jmshapir, line 58: Previously we had `/source/paper/References`. If we point only to `References` I think we will need `/source/paper` on the input path. I've given that a try above. Can you see if it allows us to compile the bibliography via `scons`?

source/lib/JMSLab/builders/build_latex.py
jmshapir, line 1: @ew487 were the changes to the builder here intentional? If not can we revert? Thanks!

jmshapir (changes requested):
@ew487 thanks!

I've flagged a couple points above. I suggest we try to converge on those before we check the errors flagged in `Actions`.

source/lib/JMSLab/builders/build_latex.py
ew487: @jmshapir Yes I think so, but @rcalvo12 would know more about that.

jmshapir: OK! I'm guessing these edits are the cause of the actions failure (see below).

Depending on the intent of these edits we can either revert them or change the testing.

@rcalvo12 let us know your thoughts thanks!

```
FAILED source/lib/JMSLab/tests/test_build_latex.py::TestBuildLateX::test_bibtex_basic - JMSLab._exception_classes.TargetNonexistenceError: The following target files do not exist after build:
    build/latex.pdf
    build/latex.log
    build/latex.blg
FAILED source/lib/JMSLab/tests/test_build_latex.py::TestBuildLateX::test_default - AttributeError: 'LatexBuilder' object has no attribute 'bibtex_system_call'
FAILED source/lib/JMSLab/tests/test_build_latex.py::TestBuildLateX::test_list_arguments - AttributeError: 'LatexBuilder' object has no attribute 'bibtex_system_call'
FAILED source/lib/JMSLab/tests/test_build_latex.py::TestBuildLateX::test_nonexistent_source - AttributeError: 'LatexBuilder' object has no attribute 'bibtex_system_call'
```

source/paper/Template.tex
jmshapir: To do:
* Look at where `scons` is putting the BBL file
* See what happens if we move references to `/source/refs/References.bib` but keep the line above as `\bibliography{References}`.

source/lib/JMSLab/builders/build_latex.py
jmshapir: Per mtg: reverting this change seems (?) to have fixed unit tests.

@rcalvo12 let's see if there's a way to reimplement the improvements you made to the builder while still maintaining testing.

rcalvo12: @jmshapir I spent about an hour trying to figure out how to reimplement the code cleanup while maintaining testing but was not able to come up with a solution. Given that these changes are mostly just for aesthetics, I don't think it is worth it to keep working on this at this point.

source/paper/Template.tex
ew487: @jmshapir I tried some of these suggestions but haven't gotten it to work so far. There was one way I did get it to work which I pushed to a new branch, https://github.com/JMSLab/Template/tree/85-testing. I actually prefer this new way but if you don't like it, I will take another shot at the suggestions.

jmshapir: @ew487 that is a great idea! 🤯 

I didn't know about the `ifthen` package but it [looks like](https://ctan.org/pkg/ifthen?lang=en) it ships with LaTeX which is nice.

I'm going to close this pull without merge and open a new one for the alternative you've proposed.

PR #87, opened 2023-11-10
Pull request for #85: Explore whether we can change file path structure for LaTeX files

jmshapir:
@rcalvo12 this pull is for an alternative to the solution we looked at in #86.

I've tested it locally and confirmed that I can compile the repo via `scons` and also compile `Template.tex` locally in my `TeX` editor.

Can you confirm the same? 

If so please approve!

Thanks.

### DIFF
--- a/source/figures/gdp_educ.tex
+++ b/source/figures/gdp_educ.tex
@@ -5,7 +5,7 @@
 \medskip{}
 \par\end{centering}
 \begin{centering}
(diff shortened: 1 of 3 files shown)


jmshapir:
@rcalvo12 probably also good to push output when you rebuild via scons just to keep logs etc up to date.

Thanks!

rcalvo12 (approved):
@jmshapir, this solution worked exactly as planned when I tested locally, so I am approving.

PR #89, opened 2023-12-15
PR for #88: Add latex template for slides

### LINKED ISSUE #88: Add LaTeX template for slides
@ew487 can we:
- [ ] add a LaTeX file with a very basic beamer shell (from a public source) for slides to [`/source/talk`](https://github.com/JMSLab/Template/tree/204def9224fc4259dad68c77a04aa190fe008291/source/talk) in which we:
   - [ ] enable the option [`\setbeameroption{hide notes}`](https://tug.ctan.org/macros/latex/contrib/beamer/doc/beameruserguide.pdf#198) by default
   - [ ] add an example in-frame speaker note to one of the slide frames
- [ ] add to the LaTeX builder a `handout` option similar to the one we have for LyX
   * under the hood, all this should need to do is switch `hide notes` to `show notes`

Thanks!

ew487:
Closes #88 

In this issue, we added a LaTeX file with a basic beamer shell for slides to [/source/talk] in which we enable the option `\setbeameroption{hide notes}` by default. We added to the LaTeX builder a handout option similar to the one we have for LyX
under the hood, which switches `hide notes` to `show notes`.


ew487:
@jmshapir I reverted the implementation to use the `note` command!

Do you want to use the same kind of input control as we did for the lyx handout?
image
I think this is causing some of the tests to fail...

source/talk/TemplateSlides.tex
```diff
@@ -0,0 +1,28 @@
...
+	\item Point One 
+	\note{Notes on Main point}
+	\begin{itemize}
+		\item Sub Point
+		\item Sub Point  
+		\note{Notes on subpoint}
+	\end{itemize}
+ 	\item Point Two
+\end{itemize}
+\end{frame}
```
jmshapir, line 26: @ew487 can we add an example transition slide between talk sections roughly along the lines of the below?

(Even better, but not essential, if these can be linked to the use of a `\section{}` designation.}

image

source/lib/JMSLab/builders/build_latex.py
jmshapir, line 1: Should we add unit testing for new functionality?

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -47,7 +50,83 @@ def add_call_args(self):
                                   os.path.normpath(self.log_file))
         self.call_args = args
         return None
-     
+    
+    
+    def check_handout(self, target, env):
```
jmshapir, line 55: @ew487 on the below, I guess maintaining parallelism with LyX is ideal if we can do it. If you see a reason why this isn't practical for the TeX builder let me know and we can figure out how best to compromise.

Thanks!

> Do you want to use the same kind of input control as we did for the lyx handout? image I think this is causing some of the tests to fail...

jmshapir (changes requested):
@ew487 thanks!

In the pull threads I replied to your question and added a few other points.

@rcalvo12 I'm adding you as reviewer both to review code and to recompile the repo. (Locally or on the cluster, as you prefer.)

Thanks!

source/lib/JMSLab/builders/build_latex.py
ew487: @jmshapir I think it would work practically! But I just wanted to check because from looking at the unit testing it seems like we're altering the expected behavior of the builder, so if someone based their project off of this template in the past, it might not work anymore?

jmshapir: @ew487 thanks! To clarify, do you mean that the change we're considering here to the LaTeX builder will not be backwards compatible, in the sense that if someone adopts the new builder in an existing repository, they may need to update the syntax they use to call the builder?

If so, that sounds very likely fine. To help me think it through, can you give me an example of syntax that would work with the current builder (i.e. the one in `main`) but not in the alternative we're considering (here in #89)?

jmshapir: To follow up here, while I think it's likely fine to change the syntax of the LaTeX builder, we probably _don't_ want to deprecate any functionality (e.g. bibtex etc.) Anyway we can figure it out once I have a better sense of the issues here thanks.

source/lib/JMSLab/tests/test_build_latex.py
```diff
@@ -13,17 +13,19 @@
 
 from ..builders.build_latex import build_latex, LatexBuilder
 from .._exception_classes import ExecCallError
+from .. import misc
```
rcalvo12, line 16: @ew487 Per spyder, this package is imported but not used. Might want to double check.

rcalvo12 (changes requested):
@ew487 Awesome work on this! Especially the way you cleaned up the latex builder a bit. Just a couple of minor points below.

@jmshapir I also recompiled the repository in 3ed0ae6.

output/talk/TemplateSlidesLatex.pdf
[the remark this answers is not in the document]
jmshapir: @rcalvo12 thanks!

When we experimented before we were [having an issue with importing EPS files](https://github.com/JMSLab/Template/issues/88#issuecomment-1857786364). It seemed to be specific to the compiler, so if it's working for you, it would be great to add a slide with [this plot](https://github.com/JMSLab/Template/blob/380e12f4d8cb94165fd9f86262894780b008be1e/output/analysis/plots/gdp_educ.eps) for illustration, thanks!

source/talk/TemplateSlides.tex
ew487: Added [here](https://github.com/JMSLab/Template/pull/89/commits/57caa68034b01a423174e09db52d50ccc68c7d8e)!

output/talk/TemplateSlidesLatex.pdf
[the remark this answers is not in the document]
rcalvo12: @ew487 @jmshapir in https://github.com/JMSLab/Template/pull/89/commits/a2be61ab86ccacb923a2dbb9d1c62a19f0eb6cd1 I went ahead and added plots to both slide templates. I didn't run into any issues compiling with scons locally and all the tests passed, so I think it worked. @ew487 would you be able to confirm?

source/lib/JMSLab/tests/test_build_latex.py
rcalvo12: @ew487 Done in https://github.com/JMSLab/Template/pull/89/commits/11a1c76e2a5668238b3bdb2fd1abd0e10d892a49. Tests ran as usual after the removal so I am resolving.

ew487: @rcalvo12 Great, thank you!

source/lib/JMSLab/builders/build_latex.py
ew487: Added in [daed0b7](https://github.com/JMSLab/Template/pull/89/commits/daed0b70b92865b7d2fa82e46833310207c00c23) and [39b7682](https://github.com/JMSLab/Template/pull/89/commits/39b7682b122a0a99aa48d664640200c951f60dd4)!

source/lib/JMSLab/builders/build_latex.py
ew487: We have not deprecated any functionality and are just requiring that the user enter target files using a specific syntax. So whereas before they are able to write `target = ["Template_tex.pdf", "Template_tex.log"], now we are enforcing that the second target file (if it exists) must be the path to the handout.

Per meeting, @jmshapir is fine with this so I will resolve the thread.

output/talk/TemplateSlidesLatex.pdf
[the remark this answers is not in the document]
ew487: @rcalvo12 Yes, it worked for me too! Thanks!

ew487:
@jmshapir @rcalvo12 Thanks for your comments--I think I've addressed all of them now! If there is anything else, please let me know.

jmshapir (approved):

rcalvo12 (approved):
@ew487 Thanks, barring the reminder posted below this looks good to me!

output/talk/TemplateSlidesLatex.pdf
rcalvo12, line 1: @ew487 Reminder to .gitignore this pdf!

ew487:
@rcalvo12 Thanks for catching that! Added to the .gitignore in [c61131e](https://github.com/JMSLab/Template/pull/89/commits/c61131e035e993fac1b125cf0e6d3919cf3da4b8) and removed in [361a934](https://github.com/JMSLab/Template/pull/89/commits/361a934601158cd07fe9580faac97fe5d55fb65f)

ew487:
Summary [here](https://github.com/JMSLab/Template/issues/88#issuecomment-1893883837)

PR #99, opened 2024-05-13
PR for #98 add [update open branches from main after completing a PR] to workflow

zhizhongpu:


jmshapir:
@zhizhongpu thanks! 

I made a few tweaks in [4c9e1e8](https://github.com/JMSLab/Template/pull/99/commits/4c9e1e8afcc392104b10cef185ea99a24a40f906). Let me know if any concerns.

Otherwise, it looks like we have some failures in macos build testing. (At least one ubuntu test has passed though.) I think it's very unlikely to be related to the change we made here, so I wonder if we need to update our syntax for github actions. 

Can you look into it?

jmshapir:
@zhizhongpu can you let me know our status here?

It's possible I may be able to ask someone else to jump in to help.

Thanks.

jmshapir:
@zhizhongpu thanks!

I'm coassigning @ew487. For now we can plan that whoever gets to this first (@zhizhongpu or @ew487) can work on it. 

I will be in touch separately about how best to prioritize this with other work.

@ew487 please see https://github.com/JMSLab/Template/pull/99#issuecomment-2108415188 for where we left off here.

jmshapir:
@zhizhongpu thanks!

> I'm not sure what the goal of these checks are, 

The goal is to increase our confidence that the builders work on a variety of systems.

> @jmshapir lmk your thoughts. 

Let's go with (2). Let me know when it's implemented in the main issue branch here, thanks!

jmshapir (approved):

Issue #102, opened 2024-09-10
Move SaveData


liaochris:
Move SaveData.py gslab-econ/gslab_python to JMSLab/TemplatePer lab meeting, use as application in https://github.com/JMSLab/Template/blob/main/source/derived/wb_clean/build.py

liaochris:
@jmshapir I need write access to the repo in order to create a development branch. Thanks!

jmshapir:
@liaochris should be done thanks!

liaochris:
In this issue, we migrated `SaveData.py` and associated dependencies from `gslab-econ/gslab_python` to `JMSLab/Template`. Here's a list of the important changes
- `SaveData.py` new home: [`source/lib/JMSLab/SaveData.py`](https://github.com/JMSLab/Template/blob/1184747f95523c4c2dcf3564ee1efc9ddf67f2d9/source/lib/JMSLab/SaveData.py)
- Implementation of `SaveData` in [`source/derived/wb_clean/build.py`](https://github.com/JMSLab/Template/blob/1184747f95523c4c2dcf3564ee1efc9ddf67f2d9/source/derived/wb_clean/build.py)
- Unit tests`test_savedata.py` and `data/data.csv` in the [`tests`](https://github.com/JMSLab/Template/tree/1184747f95523c4c2dcf3564ee1efc9ddf67f2d9/source/lib/JMSLab/tests) folder
- Updated [Github Actions workflow](https://github.com/JMSLab/Template/blob/1184747f95523c4c2dcf3564ee1efc9ddf67f2d9/.github/workflows/pytest.yml) to test Python 3.9-3.11 instead of Python 3.8-3.10

Latest version of issue branch prior to merge: 1184747f95523c4c2dcf3564ee1efc9ddf67f2d9
Branch merged to master in https://github.com/JMSLab/Template/commit/136f43826d79ebe8780d01ed26a966a130ab8a42

PR #103, opened 2024-09-13
PR for #102: Move SaveData from gslab-econ to JMSLab

liaochris:
In this pull request, I 
- moved the latest version of [`SaveData.py`](https://github.com/gslab-econ/gslab_python/blob/1524a57c2c48019ddc2af4b619045adc84a01725/gslab_misc/SaveData/SaveData.py) from `gslab-econ/gslab_python` to `JMSLab/Template`. The file now lives in [`source/lib/JMSLab/SaveData.py`](https://github.com/JMSLab/Template/blob/9cc5bb10105454654353f52981e5edc0b85ca0f5/source/lib/JMSLab/SaveData.py)
- moved the [unit tests](https://github.com/JMSLab/Template/blob/9cc5bb10105454654353f52981e5edc0b85ca0f5/source/lib/JMSLab/tests/test_savedata.py) and associated [data](https://github.com/JMSLab/Template/blob/102-move-savedata/source/lib/JMSLab/tests/data/data.csv), and tested them by running `pytest source/lib/JMSLab/tests/test_savedata.py` from the repository root. 
- amended [`source/derived/wb_clean/build.py`](https://github.com/JMSLab/Template/blob/9cc5bb10105454654353f52981e5edc0b85ca0f5/source/derived/wb_clean/build.py) to use `SaveData` 
- reran `scons`


liaochris:
@jmshapir Unit tests for python 3.8 were failing because in https://github.com/gslab-econ/gslab_python/issues/176, we added a setting that prevented "silent downcasting" when forward filling. 

Removing that setting passes all tests but resets progress made in https://github.com/gslab-econ/gslab_python/issues/176. Given that python 3.8 is outdated ([isn't receiving active bug fixes anymore](https://www.python.org/downloads/release/python-3817/)), do you think it makes sense to remove all unit tests related to python 3.8 and replace them with python 3.11 unit tests? 

FYI @zhizhongpu

jmshapir:
@liaochris thanks! I'd be fine stepping up the unit testing to python 3.11.

There's a related question about whether we want the `SaveData` unit testing integrated with our github actions workflow. I can see arguments either way, but my initial instinct is to leave the `SaveData` unit testing out of the actions workflow, which is what I believe you've done here.

I've added myself and @zhizhongpu as reviewers here; maybe we can review once you've updated the unit tests per above.

liaochris:
@jmshapir 
> I can see arguments either way, but my initial instinct is to leave the `SaveData` unit testing out of the actions workflow, which is what I believe you've done here.

The unit testing is actually part of the actions workflow, which is why I was encountering the issue with failing unit tests earlier - see the bottom of this screenshot. 

Screenshot 2024-09-13 at 11 48 56 AM


> I'd be fine stepping up the unit testing to python 3.11.

I will add unit tests for python 3.11, remote unit tests for python 3.8, and once that is done, notify you and @zhizhongpu that the PR is ready for review.

jmshapir:
@liaochris thanks! In that case is your proposal to universally disable unit testing for python 3.8 in our actions workflow? Or to create an exception for `SaveData` testing?

liaochris:
@jmshapir My proposal is to universally disable unit testing for python 3.8

jmshapir:
@liaochris thanks. Would you also disable testing for 3.9, 3.10?

I'd ideally like to preserve as much backwards-compatibility as possible.

liaochris:
@jmshapir 

> Would you also disable testing for 3.9, 3.10?

I would keep testing for those. Python 3.8 is the only version of python where the latest available version of pandas doesn't have the prevent silent downcasting" when forward filling option that's causing issues. 

Alternatively, we can remove `SaveData` tests from the github actions framework, which would allow us to keep Python 3.8 builds. 

In summary, we should decide whether we want to 
a) Remove `SaveData` tests from Github Actions (so we can keep github actions builds for python 3.8)
b) Keep `SaveData` tests in Github Actions (and remove python 3.8 from our github actions builds)

jmshapir (approved):

liaochris:
Summary comment https://github.com/JMSLab/Template/issues/102#issuecomment-2349815060

Issue #114, opened 2024-10-25
Automatically export pythonpath in SConstruct


liaochris:
Here we are updating the template to automatically export the PythonPath so that we don't have to use `sys.path.append` in every file where we import helper functions from other parent directories.

liaochris:
![Screenshot 2024-10-30 at 8 45 46 AM](https://github.com/user-attachments/assets/e19cb30e-cb9b-4657-ac76-5397d2cb9982)
Confirming that the pythonpath export is local to the scons run

PR #115, opened 2024-10-30
PR for #114: Export pythonpath in builder

liaochris:
In this issue we updated the builder and other files in the lab template to
1. remove `sys.path.append` usage
2. export the pythonpath in the builder


liaochris:
@zhizhongpu let me look into this, will tag you when resolved

liaochris:
We have two options
1. Use https://pypi.org/project/pytest-env/ to export environment variables to pytest before running
2. change GA actions to run `export PYTHONPATH=.` before pytest

I prefer the second option because it's much simpler and less intrusive (doesn't involve updating requirements)

liaochris:
@zhizhongpu i fixed this using a third solution - using sys.path in `pytest`. Since we're only using it in `pytest` and not `scons` I think it's fine but let me know what you think.

liaochris:
@zhizhongpu note also that in https://github.com/JMSLab/WorldOrder/issues/33, Joshua tested `scons` and the same SConstruct works

liaochris:
summary link https://github.com/JMSLab/Template/issues/114#issuecomment-2483505485

Issue #97, opened 2024-04-24
Automatic splitting of main text and appendix bibliographies


miikapaal:
I've found a way to create separate bibliographies for the main text and appendix so that the appendix bibliography only includes references that are not cited in the main text.

[This file](https://github.com/miikapaal/su-econ-dissertation-template/blob/main/2_chapter/main.tex), from my public dissertation template, shows the implementation which uses [`biblatex`](https://github.com/miikapaal/su-econ-dissertation-template/blob/89e5267b1c432b66c3ffc246009099bc30f05001/dissertation.tex#L7-L13C12).

It seems that the main hurdle would to be modify the LaTeX builder to use `biblatex` instead of `bibtex`.

Would there be interest in adding this feature? I'm busy putting together my dissertation right now, but I'd be happy to work on this in my spare time over the summer. Of course, any help would be appreciated!

FYI @jmshapir

miikapaal:
> @miikapaal thanks!
> 
> What I will plan to do is to keep an eye out for a live use case where we can test this out on our end. If that test is successful then we can test out adding it to the template.
> 
> Meantime we can keep this issue thread open as a reminder.
> 
> @rcalvo12 @zhizhongpu @ew487 fyi

Thanks @jmshapir, sounds good!
Just wanted to add that I've only implemented this with LaTeX; I haven't checked if it works in LyX.

jmshapir:
**Summary:** In this issue we added support for a separate appendix bibliography in `LyX`. We also incremented the `pandas` version to address a possible error. Changes brought to `main` in [d780838](https://github.com/JMSLab/Template/commit/d780838409d3f9cdcb495cb6236ada7e7583f7c6).

PR #118, opened 2025-01-04
Pull request for #97: Automatic splitting of main text and appendix bibliographies

jmshapir:
In this issue we:
* enabled multiple bibliographies in `Template.lyx`
* stepped up the LyX versions of files and harmonized their encodings

Goals for review:
* Check changes and functionality
* Recompile in `scons`


jmshapir:
@miikapaal thanks for this suggestion! Do you want to take the [review ](https://github.com/JMSLab/Template/pull/118#issue-2768938749) here?

Note that I haven't yet implemented the solution for `TeX`. I think I may have an opportunity to test that on a live project soon, after which (if successful) I can circle back to add to the template.

Thanks!

miikapaal:
> @miikapaal thanks for this suggestion! Do you want to take the [review ](https://github.com/JMSLab/Template/pull/118#issue-2768938749) here?
> 
> Note that I haven't yet implemented the solution for `TeX`. I think I may have an opportunity to test that on a live project soon, after which (if successful) I can circle back to add to the template.
> 
> Thanks!

Thanks @jmshapir! It looks like the implementation essentially follows [Section 2.6 of this doc](https://johnrhudson.me.uk/computing/Tips_tricks_and_extended_features_of_LyX_LaTeX.pdf), is that right?

I confirm that I can compile the LyX file. Note that I needed to re-install LyX (upgrade from version 2.3.5.2 to 2.4.2.1) to be able to open the template LyX file.

However, I cannot compile it with `scons` since I don't have access to `./datastore`, see screenshot. If you think testing `scons` is not important, I'm happy to approve.
Screenshot 2025-01-08 at 18 22 05

jmshapir:
@miikapaal thanks!

> Thanks @jmshapir! It looks like the implementation essentially follows [Section 2.6 of this doc](https://johnrhudson.me.uk/computing/Tips_tricks_and_extended_features_of_LyX_LaTeX.pdf), is that right?

It's similar though I think I selected a slightly different option in the LyX document settings.

> I confirm that I can compile the LyX file. 

Great.

> Note that I needed to re-install LyX (upgrade from version 2.3.5.2 to 2.4.2.1) to be able to open the template LyX file.

Makes sense. I think there was a non-backwards-compatible format change from LyX 2.3.x to LyX 2.4.x. Since LyX 2.4 is now in 2.4.2 it seems reasonable to expect users to have 2.4.x; let me know if you disagree though.

> However, I cannot compile it with `scons` since I don't have access to `./datastore`, see screenshot. If you think testing `scons` is not important, I'm happy to approve.

Can you access it [here](https://drive.google.com/drive/folders/1b3UeaAYbtOk67qyniBGLLqkVIHE4xSvy)? If so, you should be able to download a local copy to the right location (or symlink it).

miikapaal:
Thanks @jmshapir!

> @miikapaal thanks!
> 
> > Thanks @jmshapir! It looks like the implementation essentially follows [Section 2.6 of this doc](https://johnrhudson.me.uk/computing/Tips_tricks_and_extended_features_of_LyX_LaTeX.pdf), is that right?
> 
> It's similar though I think I selected a slightly different option in the LyX document settings.
> 
> > I confirm that I can compile the LyX file.
> 
> Great.
> 
> > Note that I needed to re-install LyX (upgrade from version 2.3.5.2 to 2.4.2.1) to be able to open the template LyX file.
> 
> Makes sense. I think there was a non-backwards-compatible format change from LyX 2.3.x to LyX 2.4.x. Since LyX 2.4 is now in 2.4.2 it seems reasonable to expect users to have 2.4.x; let me know if you disagree though.

Yes, I agree. Also, the error message I got when trying to open the file clearly indicated my LyX version was outdated.

> 
> > However, I cannot compile it with `scons` since I don't have access to `./datastore`, see screenshot. If you think testing `scons` is not important, I'm happy to approve.
> 
> Can you access it [here](https://drive.google.com/drive/folders/1b3UeaAYbtOk67qyniBGLLqkVIHE4xSvy)? If so, you should be able to download a local copy to the right location (or symlink it).

I was just about to write that I added a local copy of `datastore` as per the instructions in the [README](https://github.com/JMSLab/Template/blob/main/README.md). But I still can't compile due to some error with `source/lib/SaveData.py` in `source/derived/wb_clean/build.py`, see screenshot. So the issue is not related to the changes made in #97.

Screenshot 2025-01-08 at 18 49 14

liaochris:
@miikapaal 
Two points
1. What is your pandas version? You can find it by running 
```
import pandas as pd
pd.__version__
```

2. I believe it should be possible to fix the error by removing the `copy = False` argument from `SaveData.py` and it won't change the results of the function call. Could you test this to confirm on your end? This is the case on my machine. I'm not sure if this is the source of your error but the [copy argument will be deprecated in future versions of pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.infer_objects.html) so it seems reasonable to me to pre-empt that by removing it, especially if
	a) It's causing errors for other users
	b) `copy = True` or `copy = False` doesn't change the outcome of the relevant function call, which is to summarize some data. 
	
FYI @jmshapir

miikapaal:
Thanks @jmshapir and @liaochris!

> @miikapaal Two points
> 
> 1. What is your pandas version? You can find it by running
> 
> ```
> import pandas as pd
> pd.__version__
> ```
> 

It's 1.3.2.

> 2. I believe it should be possible to fix the error by removing the `copy = False` argument from `SaveData.py` and it won't change the results of the function call. Could you test this to confirm on your end? This is the case on my machine. I'm not sure if this is the source of your error but the [copy argument will be deprecated in future versions of pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.infer_objects.html) so it seems reasonable to me to pre-empt that by removing it, especially if
>    a) It's causing errors for other users
>    b) `copy = True` or `copy = False` doesn't change the outcome of the relevant function call, which is to summarize some data.
> 

@liaochris I can confirm that compiling with `scons` works on my machine (Mac) after I dropped the `copy = False` argument from `SaveData.py`. So if you believe dropping the argument does not cause problems in other settings, then doing that makes sense. Maybe you want to first test that this is indeed the case [in the `gslab_python` repo](https://github.com/gslab-econ/gslab_python/blob/master/gslab_misc/SaveData/SaveData.py) by running the unit tests?

@jmshapir I wanted to note that your implementation actually _does not_ filter out references that already appear in the main text from the appendix bibliography, see screenshot. If this is fine with you, then I'm happy to approve!

Screenshot 2025-01-09 at 16 57 18

jmshapir:
@miikapaal 

> @jmshapir I wanted to note that your implementation actually _does not_ filter out references that already appear in the main text from the appendix bibliography, see screenshot. If this is fine with you, then I'm happy to approve!

Yes, I'm fine with that, thanks! Either way let's wait to merge until we've resolved the point about pandas.

@liaochris 

> It's 1.3.2.

Per our chat today it sounds like we may want to require a more recent pandas in [`requirements.txt`](https://github.com/JMSLab/Template/blob/4540a8bf90966d7cab2588750fbf84ce75e9e802/source/lib/requirements.txt)?

Thanks both!

liaochris:
@miikapaal 
Per discussion with @jmshapir, it seems like a better solution to this would be to upgrade your pandas version to pandas 2.0. We will also update the `source/lib/requirements.txt` on our end. 

In [pandas 1.3](https://pandas.pydata.org/pandas-docs/version/1.3/reference/api/pandas.DataFrame.infer_objects.html), the `infer_objects` command does not have the `copy` attribute as it was introduced in [pandas 2.0](https://pandas.pydata.org/pandas-docs/version/2.0/reference/api/pandas.DataFrame.infer_objects.html)

miikapaal (approved):
> @miikapaal Per discussion with @jmshapir, it seems like a better solution to this would be to upgrade your pandas version to pandas 2.0. We will also update the `source/lib/requirements.txt` on our end.
> 
> In [pandas 1.3](https://pandas.pydata.org/pandas-docs/version/1.3/reference/api/pandas.DataFrame.infer_objects.html), the `infer_objects` command does not have the `copy` attribute as it was introduced in [pandas 2.0](https://pandas.pydata.org/pandas-docs/version/2.0/reference/api/pandas.DataFrame.infer_objects.html)
> 
> Could you upgrade your pandas and retest? Thanks and sorry for the back and forth!

Thanks @liaochris @jmshapir ! After updating `pandas` I could successfully compile the PDF with `scons` so I'm approving. Feel free to merge to `main`.

liaochris:
Thanks @miikapaal 

@jmshapir I upgraded the `source/lib/requirements.txt` file and all tests are passing.

jmshapir:
@miikapaal @liaochris thanks both!

Could one of you push your successful `scons` compile? (Recall that we intentionally [ignore](https://github.com/JMSLab/Template/blob/136f43826d79ebe8780d01ed26a966a130ab8a42/.gitignore#L29-L33) a number of files because LFS-tracked files can't "ship" with a template.)

After that if everything looks good I will merge thanks.

Issue #119, opened 2025-01-13
Modify LaTeX template to include separate appendix bibliography


jmshapir:
The goal of this issue is to implement a second appendix bibliography in our [LaTeX template](https://github.com/JMSLab/Template/blob/204def9224fc4259dad68c77a04aa190fe008291/source/paper/Template.tex) that parallels the LyX implementation in #97.

jmshapir:
@miikapaal would you be interested in working on this? I'm curious if we can use `multibib` and `btUnit` similar to the implementation in #97. That won't initially have the slick feature you [included in your dissertation template](https://github.com/miikapaal/su-econ-dissertation-template/blob/8f518a2919351224fec00896cb17bc2630594c2f/2_chapter/main.tex#L3-L6) of automatically excluding main-text references from the bibliography, but I think it's ok to leave that out for a first pass.

If you don't have bandwidth for this, just let us know and we can see if someone else can jump in. Thanks!

jmshapir:
@miikapaal thanks and granted! Please use it wisely. :-)

miikapaal:
@liaochris @jmshapir 
I just started the work on this. When using multiple bibliographies, it looks like the `bibtex` executable needs to be called once for each bibliography, see the separate `.aux` files below from the compilation with LyX.

https://github.com/JMSLab/Template/blob/d780838409d3f9cdcb495cb6236ada7e7583f7c6/output/paper/sconscript.log#L10-L11
https://github.com/JMSLab/Template/blob/d780838409d3f9cdcb495cb6236ada7e7583f7c6/output/paper/sconscript.log#L14-L15

I'll check if the LaTeX Scons builder currently accommodates this type of behavior.

miikapaal:
Following up on https://github.com/JMSLab/Template/issues/119#issuecomment-2589872908, I just confirmed that a simple script like this is able to compile the PDF identical to the one compiled by the LyX builder:
```
#!/usr/bin/python

import os, subprocess, sys

os.chdir(os.path.join("source","paper"))

f = os.path.join("Template")

commands = [
    ["pdflatex", f + ".tex"],
    ["bibtex",   f + ".1" + ".aux"],
    ["bibtex",   f + ".2" + ".aux"],
    ["pdflatex", f + ".tex"],
    ["pdflatex", f + ".tex"],
]

for c in commands:
    subprocess.call(c)
```

I'll check how to adapt the LaTeX builder,with as few changes as possible.

miikapaal:
[...]

PR #122, opened 2025-02-03
Pull request for #119: Modify LaTeX template to include separate appendix bibliography

miikapaal:
@jmshapir @liaochris Could you take the review here?

I edited the latex builder by adding a `multibib` option, see [here](https://github.com/JMSLab/Template/issues/119#issuecomment-2601024162). I also added a related unit test.

Tasks:
- Review changes to [`build_latex.py`](https://github.com/JMSLab/Template/blob/aa47cbc9819b26f079859a50a62af1215f4c4ccf/source/lib/JMSLab/builders/build_latex.py) @jmshapir @liaochris 
- Review changes to [`test_build_latex.py`](https://github.com/JMSLab/Template/blob/aa47cbc9819b26f079859a50a62af1215f4c4ccf/source/lib/JMSLab/tests/test_build_latex.py) @jmshapir @liaochris 
- Confirm that all unit tests pass @liaochris 
- Confirm that compiling `output/paper/TemplateTex.pdf` works and commit logs from a build. @miikapaal or @liaochris 

Thanks!


jmshapir:
@miikapaal the automated unit tests are failing. (Screenshot below.) Can you look into that?

image

jmshapir (changes requested):
@miikapaal I went over the code and don't have any concerns except for the failed Actions checks.

Thanks!

miikapaal:
> @miikapaal the automated unit tests are failing. (Screenshot below.) Can you look into that?
> 
> image

Thanks @jmshapir. The reason in all cases is that the `pdflatex` executable is not found, see e.g. [here](https://github.com/JMSLab/Template/actions/runs/13123072697/job/36613372484?pr=122#step:5:126).

I'm assuming that all unit tests should be using the "system mock" approach (that is used for the other unit tests) since this error does not then occur? If that's right, then I'll see how to modify the new `multibib` unit test to also use that approach. Hopefully that solves this.

jmshapir:
@miikapaal thanks! Let's assume that's the approach we want and @liaochris can let us know if we should course-correct. Thanks.

liaochris:
Thanks for your work on this @miikapaal ! I agree with what @jmshapir said below and I will review the code once the checks pass successfully.

> @miikapaal thanks! Let's assume that's the approach we want and @liaochris can let us know if we should course-correct. Thanks.

jmshapir:
@miikapaal just checking in here? No special urgency but it's probably good to maintain momentum esp. now that we're so close to a working solution here. Thanks.

miikapaal:
> @miikapaal just checking in here? No special urgency but it's probably good to maintain momentum esp. now that we're so close to a working solution here. Thanks.

@jmshapir I am very, very sorry for the lack of progress here. I have needed to push some research projects intensely, as well as deal with a rather tedious data application process. I have simply forgotten to finish this PR. I will work on this today and implement [this fix](https://github.com/JMSLab/Template/pull/122#issuecomment-2634174913). Again, apologies for the delay.

miikapaal:
@liaochris @jmshapir 
I would really appreciate some help/suggestions with adding a unit test for the `multibib` feature that actually works.

In https://github.com/JMSLab/Template/pull/122/commits/adce8e067a9432780ab295a73d88dfcd96628429
I revised the implementation of `multibib` so that now:
- we have a function that parses the source `.tex` file, 
- counts the number of bibliography calls, and then 
- creates a list of `.aux` files (`target`.`n` for n = 1,...,[number of bibliography calls]) on which `bibtex` should be called

The advantage of this is that, [unlike previously](https://github.com/JMSLab/Template/issues/119#issuecomment-2601024162), we do not need to first run `pdflatex` in order to figure out what the list of `.aux` should be.

My thinking was that the revised version would allow using the `system mock` approach (see https://github.com/JMSLab/Template/pull/122#issuecomment-2634174913) when conducting unit tests. This is useful since it appears the GitHub Actions feature this repo is using for unit testing does not have `pdflatex` as an available executable.

However, even after some small edits to [the new unit test](https://github.com/JMSLab/Template/pull/122/commits/27ad822f456df9a4c20d1dafb9bb4006cd295c9c), the unit tests don't pass. (They do not pass locally on my machine, either.) The code architecture for unit testing is quite complicated, so I don't know how to fix this.

My hunch is that we might need to edit one or more of the functions [`standard_test`](https://github.com/JMSLab/Template/blob/311dbb10d7093a2153ecbecb71a1ec7ea71a622f/source/lib/JMSLab/tests/_test_helpers.py#L171), [`latex_side_effect`](https://github.com/JMSLab/Template/blob/311dbb10d7093a2153ecbecb71a1ec7ea71a622f/source/lib/JMSLab/tests/_side_effects.py#L190), and [`bibtex_side_effect`](https://github.com/JMSLab/Template/blob/311dbb10d7093a2153ecbecb71a1ec7ea71a622f/source/lib/JMSLab/tests/_side_effects.py#L243) to solve this. Maybe the mocked machine does not create the `.aux` files which `pdflatex` creates as part of compiling the target PDF.

jmshapir:
@mcaceresb any chance you'd have a few minutes to look into the roadblocks @miikapaal is encountering with updating unit tests here?

@miikapaal @mcaceresb fyi I think if we had to ask the user to specify the number of bibliographies (where the default is, say, one), rather than asking the code to figure it out automatically, that would probably be fine, at least for our initial implementation of this feature.

Thanks!

miikapaal:
> @miikapaal @mcaceresb fyi I think if we had to ask the user to specify the number of bibliographies (where the default is, say, one), rather than asking the code to figure it out automatically, that would probably be fine, at least for our initial implementation of this feature.

I don't think adding such a feature is necessary. I think we should just ensure that the mock machine can parse the source `.tex` file and create dummy versions of the necessary `.aux` files in the target folder.

Atm it seemed to me that the mock machine does the `bibtex` part assuming that the `.aux` file is called `[target].aux`, whereas with `multibib` it has the structure `target.[1-9].aux`.

mcaceresb:
@miikapaal I'm not sure why the tests on github are failing, but locally all tests run for me once I add [`input/test_multibib.tex`](https://github.com/JMSLab/Template/blob/27ad822f456df9a4c20d1dafb9bb4006cd295c9c/source/lib/JMSLab/tests/test_build_latex.py#L98) to [the list of existing files for the mock test](https://github.com/JMSLab/Template/blob/27ad822f456df9a4c20d1dafb9bb4006cd295c9c/source/lib/JMSLab/tests/_side_effects.py#L233). Does that fix it for you as well?

miikapaal:
> @miikapaal I'm not sure why the tests on github are failing, but locally all tests run for me once I add [`input/test_multibib.tex`](https://github.com/JMSLab/Template/blob/27ad822f456df9a4c20d1dafb9bb4006cd295c9c/source/lib/JMSLab/tests/test_build_latex.py#L98) to [the list of existing files for the mock test](https://github.com/JMSLab/Template/blob/27ad822f456df9a4c20d1dafb9bb4006cd295c9c/source/lib/JMSLab/tests/_side_effects.py#L233). Does that fix it for you as well?

Thank you so much, @mcaceresb! This solved the problem for me as well. I implemented this in https://github.com/JMSLab/Template/pull/122/commits/fcda4f0bb98602dc3d067259b2b92a3ac0d9f2dc. Let's see if the github tests pass now as well...

jmshapir (approved):
@miikapaal @mcaceresb thanks!

Looks good to me but let's give @liaochris a chance to complete a review.

miikapaal:
> @miikapaal @mcaceresb thanks!
> 
> Looks good to me but let's give @liaochris a chance to complete a review.

Thanks @jmshapir!

@liaochris After you've finished your review, could you also build `output/paper/TemplateTex.pdf` with SCons and commit logs from the build` (see https://github.com/JMSLab/Template/pull/122#issue-2828551546)? Thanks!

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -152,57 +153,109 @@ def check_bib(self, source):
...
         return None
 
+    def count_bibsections(self, tex_file):
+        """Count the number of \\begin{btSect}...\\end{btSect} blocks."""
+        with open(tex_file, 'r', encoding='utf-8') as f:
+            content = f.read()
+
+        btsect_blocks = re.findall(r'\\begin\{btSect\}.*?\\end\{btSect\}', content, re.DOTALL)
+        return len(btsect_blocks)
+
```
liaochris, line 163: @miikapaal I've never seen `btSect` used for bibliography references, so I'm not sure this is a superior approach to just asking the user how many bibliographies there are. We didn't use `btSect` in the original `Template.tex`.

FYI @jmshapir

source/lib/JMSLab/builders/build_latex.py
```diff
@@ -152,57 +153,109 @@ def check_bib(self, source):
...
             raise_system_call_exception = False
             try:
                 subprocess.check_output(self.handout_call, shell = True, stderr = subprocess.STDOUT)
                 if self.checked_bib:
                     self.bibtex_executable  = get_executable('bibtex')
-                    self.bibtex_system_call = '%s %s' % (self.bibtex_executable, self.out_name)
+                    self.check_multibib(target, env)
+                    if self.checked_multibib:
+                        for f in self.aux_files:
+                            time.sleep(3) # Wait a bit to make sure all .aux files have been created
```
liaochris, line 227: What if we change this to a while loop that loops until all aux files have been created, to eliminate any possibility of breaking?

source/lib/JMSLab/builders/build_latex.py
miikapaal: @liaochris
You can read [here](https://www.overleaf.com/learn/latex/Questions/Creating_multiple_bibliographies_in_the_same_document#Packages_for_BibTeX) on how to include multiple bibliographies in the same document. One thing you'll notice that the default method of using `\bibliography{...}` does not accommodate this behavior.

IMO the advantage of the current approach is that it works regardless of how many bibliographies one wants to include in the document. E.g.  if you want to split the Appendix bibliography into two (e.g., one for "main" appendix" and one for online appendix), then all you need to do is add an additional `btSect` environment in the source `.tex` file -- you don't need to change anything e.g. the `SConscript` files.

If you're fine with this, feel free to resolve.

FYI @jmshapir

source/lib/JMSLab/builders/build_latex.py
miikapaal: This is a good point.

I tried this but noticed then that the unit tests stall when trying to the new test. It might be because the mock system does not actually create the `.aux` files? 

So what I ended up doing in [046e202](https://github.com/JMSLab/Template/pull/122/commits/046e2029862d45f519e7e35b3bb9e1c2a45a7ad7) was to run the 3 sec wait only if the `.aux` has not yet been created. I also commited logs from a SCons build (cf. https://github.com/JMSLab/Template/pull/122#issuecomment-2769846700).

I would suggest we wrap up here and revisit this if it ever ends up being a problem. If you disagree, feel free to try to implement your original suggestion.

FYI @jmshapir

liaochris: I think we can wrap up here.

source/lib/JMSLab/builders/build_latex.py
liaochris: Thanks @miikapaal! That makes sense. One followup question - do we want to also include in the template an example `Template.tex` with no multibib functionality, so it's clear how to implement one bibliography? Or do you think the examples in the tes cases are sufficiently clear. If we do so, we'd probably want to rename the multibib version to `Template_multibib.tex`?

@jmshapir Our existing tests actually check (and successfully run) because the `scons` command for compiling a draft with multiple bibliographies is slightly different from the `scons` command for compiling a draft with a single bibliography!

jmshapir: @liaochris for unit testing purposes I'm happy to have a TeX file with a single bibliography, but for the main template, I think it's simpler if we just have one `Template.tex`. And for that `Template.tex`, I think we do want separate main and appendix bibliographies.

Thanks!

liaochris: Thanks @jmshapir, I've reverted https://github.com/JMSLab/Template/commit/bf15f7b77b8577302597508d5ad17c310c0ac830.

jmshapir: @liaochris thanks! What remains to do here in this thread?

liaochris (approved):
Great work @miikapaal and everyone else! Looking forward to having this functionality in the template.

source/lib/JMSLab/builders/build_latex.py
miikapaal: Thanks both! I did [this](https://github.com/JMSLab/Template/pull/122#discussion_r2025822499) already in https://github.com/JMSLab/Template/pull/122/commits/046e2029862d45f519e7e35b3bb9e1c2a45a7ad7 so we're good to close the PR. I'll go ahead and do that. :-)

Issue #123, opened 2025-04-23
Add script that removes eps dates


liaochris:
We now have two projects that use matplotlib to create eps files, which include date of creation information in the source code. In this issue, we will add a helper script that allows one to remove dates from eps file source code.

liaochris:
In this issue we added a function that removes date + version information for EPS files produced by matplotlib. We also added an example usecase. 

[Latest version of branch](https://github.com/JMSLab/Template/tree/f872f3f20b0bfeca4cb96787f96c6142d20492d4)
Branch merged in https://github.com/JMSLab/Template/commit/e1fccecbf3b9dfc1c2479912cf3315cb9e6f9fe5

PR #124, opened 2025-04-23
PR for #123: Add date removal function for EPS 

liaochris:
- Modified matlab script to only output the png file
- Created an equivalent python script that creates the eps file and is a usecase for `remove_eps_dates`


source/analysis/plots/makeplot.m
jmshapir, line 1: @liaochris my instinct is that we should ask both the matlab script and the python script to output both EPS and PNG. That way, we're communicating how we usually output graphics files.

source/lib/JMSLab/remove_eps_dates.py
jmshapir, line 1: If it wouldn't conflict with `matplotlib` TOU, would we also want to remove MatPlotLib version information, to avoid diffs due to minor version changes?

jmshapir (changes requested):

source/analysis/plots/makeplot.m
liaochris: @jmshapir Got it. `scons` doesn't work well when the same output is produced by two different scripts. Let me know which of the options you prefer or if you have in mind something else:
1) Have both scripts produce EPS + PNG, and only include one of them in the DAG
2) Rename the script outputs so that it's specific to the script

I think the second might be slightly preferable to the first, but let me know what you think!

jmshapir: @liaochris thanks! I like (2).

source/lib/JMSLab/remove_eps_dates.py
liaochris: @jmshapir I don't see anything on the matplotlib TOU that requires us to cite the specific version of matplotlib that's being used. The only additional information we need to provide when I add the version removal feature is to note in the function documentation what information we are removing (as we are already doing). 

> 3. In the event Licensee prepares a derivative work that is based on or
incorporates matplotlib or any part thereof, and wants to
make the derivative work available to others as provided herein, then
Licensee hereby agrees to include in any such work a brief summary of
the changes made to matplotlib .

liaochris: We now remove version info, see first few lines of
- [`gdp_educ_py.eps`](https://github.com/JMSLab/Template/pull/124/files#diff-0c9cb59bc492857c4480c62d1e6b3ceb8571771fab97bb2fd23a6ae62aae8913)
- and the [latest code](https://github.com/JMSLab/Template/pull/124/files#diff-c8f9619f107f223d0cd1229d2764ecb5d1235926e718a1c83e2a32e8e971ba69)

.gitignore
```diff
@@ -26,7 +26,7 @@
 *.synctex.gz
 
 # git-lfs files cannot be included in a template repository
-output/analysis/plots/gdp_educ.png
+output/analysis/plots/*.png
```
jmshapir, line 29: I think it's better to enumerate files here so that when users build off of the template they're not unintentionally gitignoring whole file categories in a folder.

output/analysis/plots/gdp_educ_py.eps
jmshapir, line 1: Instead of using the suffix `_py`, what if have the python flavor of the plot switch the axes relative to the matlab flavor, so we can then call this file `educ_gdp.eps`? (And likewise for the PNG counterpart.)

jmshapir (changes requested):
@liaochris thanks! It's looking very good to me, just a few small points to tidy up.

.gitignore
liaochris: Change made in ef78fe5eaad353c4068668622bdc35f2bc627f07

output/analysis/plots/gdp_educ_py.eps
liaochris: Change made in 30805e5d3ee5661a3a78707e779c26852d690f82

liaochris:
@zhizhongpu Just letting you know about this change

output/analysis/plots/gdp_educ_py.eps
jmshapir: @liaochris thanks. Do we also need to swap the axis titles in the python flavor so they match what we're plotting?

liaochris: Thanks, I made that adjustment in f872f3f20b0bfeca4cb96787f96c6142d20492d4.

jmshapir (approved):

Issue #127, opened 2025-06-11
Add an illustration of autofill


jmshapir:
The goal is to make it easier for a user to see how we autofill values in papers.

The steps are:
- [ ] Add script `topgdp_value` to [`source/analysis/topgdp`](https://github.com/JMSLab/Template/tree/71d716f1309a2986404876574be4dc8699687391/source/analysis/top_gdp) that uses [`autofill`](https://github.com/JMSLab/Template/blob/71d716f1309a2986404876574be4dc8699687391/source/lib/JMSLab/autofill.py) to output `output/analysis/top_gdp/top_gdp.tex` containing a single TeX macro with autofilled value
- [ ] Update the LyX and TeX flavors of the [draft](https://github.com/JMSLab/Template/tree/71d716f1309a2986404876574be4dc8699687391/source/paper) to use `output/analysis/top_gdp/top_gdp.tex` as input, and to reference the macro as an illustration

jmshapir:
@ben-workman are you up for tackling this one? If so I will elevate your privileges and assign you. Thanks!

ben-workman:
@jmshapir Hi Jesse, yes, happy to take this one!

jmshapir:
@ben-workman great thanks! I've re-invited you with write permissions. Please accept the [invite](https://github.com/JMSLab/Template/invitations) (which should also be in your e-mail) and you should be all set thanks.

jmshapir:
@ben-workman thanks for those updates! If you are done making changes can you open a pull with @jmshapir as reviewer? I can then add other reviewers as needed. Thanks.

ben-workman:
Closing this issue as it continues in its PR #128.

PR #128, opened 2025-06-24
Pull Request for #127: Add an illustration of autofill

ben-workman:
In this issue, we (1) added a Python script that uses [autofill](https://github.com/JMSLab/Template/blob/71d716f1309a2986404876574be4dc8699687391/source/lib/JMSLab/autofill.py) to output a TeX macro with an autofilled value and (2) modified the LyX and TeX versions of the draft to reference the created macro.


jmshapir:
@liaochris @zhizhongpu can one of you please provide a second review? (The other one can self-unassign.)

Thanks!

jmshapir:
@ben-workman looks like you are clear to squash-merge and wrap up when ready thanks!

PR #151, opened 2026-04-06
PR for #150: Fix bugs encountered while onboarding actions

### LINKED ISSUE #150: Fix bugs in github actions
1. Explicitly specify `contents: read` permissions in `checks.yml`

liaochris:


zhizhongpu (approved):
@liaochris 968b1e8af3254a712c9df6b9e1a5ff98e3a13efd made a readability tweak

zhizhongpu:
@jmshapir since you're added as a reviewer, here's the rationale of this modification:

- we discovered that in practice, we sometimes need to make exceptions for the scons dag check, so we've built an intuitive way to specify such exceptions
- we discovered that the Actions wouldn't work in private repos without read permission, so we gave them the read permission

.github/checks/sconscript_exceptions.py
jmshapir, line 1: Can we make this a YAML or something similar that expresses that it's a user-directed setting?

jmshapir (changes requested):

.github/checks/sconscript_exceptions.py
zhizhongpu: - Implemented by f52d0358ff7257df6a91a46d7a561c2255e9a0d3 in TOML 
- I tried a YAML version in 0a54d949b07672a7c015ac1cb3f88e354da1470f but the cloud Python set-up doesn't have `yaml` package so I switched to TOML because `tomllib` is native to Python [since 3.11](https://peps.python.org/pep-0680/).

zhizhongpu: TOML solution tested in a series of commits before fecf7c0cbe14afc0970b5097a17a9261d9c3b8dd. fecf7c0cbe14afc0970b5097a17a9261d9c3b8dd removed the tests. 

@jmshapir can resolve here if the solution looks good to you 

fyi @liaochris

zhizhongpu: > * the cloud Python set-up doesn't have `yaml` package

to clarify for @liaochris / if you're into the technical details: if we want `yaml` we'd have add the below to the checks every time we run the checks, which is unnecessarily complicating

https://github.com/JMSLab/Template/blob/fecf7c0cbe14afc0970b5097a17a9261d9c3b8dd/.github/workflows/pytest.yml#L32

.github/checks/sconscript_exceptions.toml
```diff
@@ -0,0 +1,7 @@
+[excluded_files]
+# "source/some/file.R" = "reason for exclusion"
+
+[settings]
```
jmshapir: Could we have a [skip_dirs] section structured similarly to the [excluded_files] section?

jmshapir (changes requested):
TOML works for me, just had a question about implementation/syntax.

.github/checks/sconscript_exceptions.toml
zhizhongpu: makes sense! ab955e9e2066d9027f8ecf670885132b7c2a3cbc

jmshapir: Thanks. Syntax looks different between excluded files and skip_dirs because excluded_files doesn't have a variable declaration as in 

```
excluded_files = [
      "source/some/file.R",
]
```

Is that difference intentional?

zhizhongpu: thanks @jmshapir. I reflected further and realized that we can unify them using regex + using comments to document reasons. See a more concise implementation in 2196f95b23823aee7a24327142ae8ba3cb630776

jmshapir: Nice, I think that's much clearer.

jmshapir (approved):

.github/checks/sconscript_exceptions.toml
zhizhongpu: fyi @liaochris

zhizhongpu:
@liaochris is there anything more to do here?

PR #156, opened 2026-04-24
PR for #148: Improvements to github actions workflow

liaochris:


liaochris:
@zhizhongpu I added an issue subfolder after opening the pull, and I don't see the expected behavior (being unable to merge). I don't think the bypass list has been set up either?

I think the behavior I'm testing (adding an issue subfolder in the branch) is not what the test was design to account for but is something should still account for, as it's a scenario that could occur during the normal course of development.

zhizhongpu:
@jmshapir I think we've reached a stage where it's efficient to set up the ruleset for branch protection

Please follow the steps below

a. In the repo setting for Template, add a new rule set of arbitrary name

image

b. Where I'm marking in the screenshots with a numeral, change the marked setting:

- 1: Set to "Active"
- 2: This is the bypass list. Please choose the appropriate set of users that includes @liaochris @zhizhongpu @jmshapir
- 3: Protection for `main`
- 4: Add a new requirement for status check "check-issue"
- 5: "Save changes"

image

image

image

Once these settings are applied, I expect to see the merge soft-blocked by the following interface here https://github.com/JMSLab/Template/issues/148#issuecomment-4304652794

jmshapir:
@zhizhongpu thanks. Below is the `json` for the rule I implemented. I did not include any other branch protections on this rule because we already have those in place via another rule. Let me know if any concerns.
[no-issue-in-main.json](https://github.com/user-attachments/files/27086956/no-issue-in-main.json)

zhizhongpu:
> @zhizhongpu thanks. Below is the `json` for the rule I implemented. I did not include any other branch protections on this rule because we already have those in place via another rule. Let me know if any concerns. [no-issue-in-main.json](https://github.com/user-attachments/files/27086956/no-issue-in-main.json)

thanks @jmshapir. I think this is 

1. exactly what I did too. Here's a [diff](https://www.diffchecker.com/Y6MMErJj/) of our json files.
2. working as expected

@liaochris could you see if you see this interface?

image

If so, could you test the following two:

1. that removing the issue folder in a commit titled [run-actions-all] or [run-actions-issue] runs the check
2. that the check passes once the issue folder is removed

I expect these two tests to pass. I'm approving the PR so if you approve too, ~you could merge directly~. If we need to test the comment trigger, we'd need that tested in a new PR.

[edit: perhaps wait for @jmshapir's review before merging]

zhizhongpu (approved):

jmshapir (approved):

liaochris:
@zhizhongpu I don't see the interface you're referencing in https://github.com/JMSLab/Template/pull/156#issuecomment-4320374966, and the `issue` subfolder is still on the branch. This is the interface I see. 
Screenshot 2026-05-08 at 4 10 32 PM

I'm guessing https://github.com/JMSLab/Template/pull/156/commits/63b5df864a76fde10e3beba53c3212b1e3c47a30 is affecting the test? Should we guard against that type of behavior?

liaochris:
@jmshapir 

I wanted to confirm one more thing before I merge the PR to make sure the current functionality - particularly regarding the issue subfolder checking test - is what we're aiming for. 

The issue-subfolder check is skipped by default and only runs when commits include specific trigger messages. When it's skipped (not failed), GitHub shows a neutral status rather than a red X — so a previously-flagged problem can appear to "clear" simply because a new commit ran other tests but didn't re-trigger the issue check.

This is especially relevant in Template because every commit triggers the full lab unit tests. Any follow-up commit that doesn't also trigger the issue subfolder test will show that check as skipped, making it easy to lose track of a pre-existing problem. That explains the issue I saw in https://github.com/JMSLab/Template/pull/156#issuecomment-4409506236. 

I think this is acceptable as long as we make it a habit to comment/commit `/run-actions-all` before merging. That will re-run all tests including the issue subfolder check, so any lingering problems will show up on main. Does this workflow sound good to you?

Note: per discussion with @zhizhongpu he mentioned that if we make commenting/committing `run-actions-all` a habit pre-merge, we should formalize this in our [workflows doc](https://github.com/JMSLab/Template/blob/main/docs/workflow.md).

zhizhongpu:
@liaochris thanks for posting. 

Two additional points fyi @jmshapir:

1. I previously suggested that one alternative is to have the issue check triggered by every commit, and the cost in private repos shouldn't be high (40 commits per PR * 2 seconds to run the issue check per commit should bound the total runtime to be <7 minutes for every PR we close), but perhaps it'd be annoying to receive a "failed run" notification for every commit.
2. Suppose we keep the current implementation, then the only case where we'd experience an undesirable outcome in a private repo would be where, starting from an initial position of `issue/` check failing, a user runs a specific, non-issue-folder test f"run-actions/{A}" with A not in {"issue", "all"}. Doing so will change the status of the "fail" flag to "skipped". To avoid this undesirable outcome, could we simply make the issue folder check run to be also triggered by command f"run-actions/{A}" for all A?

jmshapir:
Thanks both. I understand the limitation of the current functionality that you're highlighting. I think it's fine to merge and if we find that this behavior is limiting the usefulness of the issue folder check we can always revisit.

zhizhongpu:
> I think it's fine to merge

Thanks @jmshapir. Fine with me too. @liaochris if that plan sounds good to you could you merge? 

Then I think our next step should be to propagate these changes to ██████████?

PR #157, opened 2026-04-27
Pull request for #154: Batch of updates to production process

### LINKED ISSUE #154: Batch of updates to production process

jmshapir:
This pull brings in a variety of updates to the production process, mostly aimed at incorporating LLMs into the standard workflow.

### DIFF
--- a/docs/production.md
+++ b/docs/production.md
@@ -24,6 +24,7 @@ Production encompasses the steps we take before we circulate or submit a paper d
    * Avoid compiling on subissue branches unless necessary.
    * When merging a subissue branch, merge back to the main production branch and update other open production branches.
    * Skip review of pulls for tasks where work has already been reviewed by multiple RAs, except where PI review is requested below.
+   * It's often efficient to begin work by asking an LLM to perform the task. From there you can prune the LLM's suggestions, supplement them, or both.
 
 * When producing task deliverables:
    * Post them in a comment in the relevant subissue and tag the PI.
@@ -43,7 +44,7 @@ Production encompasses the steps we take before we circulate or submit a paper d
    
   | Expected Hours | 50% Confidence Interval | # of RAs |
   | -------------- | ----------------------- | -------- |
-  | 1              | (0.5 - 5)               | 1        |
+  | 1              | (0.5 - 3)               | 1        |
 
 #### Goals 
 
--- a/docs/production.md
+++ b/docs/production.md
@@ -129,11 +129,7 @@ Production encompasses the steps we take before we circulate or submit a paper d
 #### Goals
 
   *  Main statements made in paper about definitions of samples and variables, and other methodological choices, are consistent with code.
-  *  Checking every variable and sample definition can for some projects take a large amount of time. Accordingly:
-      * Ask an LLM to take a careful first pass. (It can be helpful to break up the main pieces of the analysis and ask separately about each).
-      * Prune the LLM's comments.
-      * Spot-check yourself, focusing especially on the main analysis and core specification(s).
-      * Limit to a day of person time unless a PI suggests otherwise.
+  *  Limit to a day of person time unless a PI suggests otherwise, and when there are tradeoffs, focus especially o the main analysis and core specification(s), as opposed to sensitivity and supplemental analyses.
 
 #### Deliverables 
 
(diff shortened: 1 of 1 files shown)


docs/production.md
simonessigaberg, line 115: @jmshapir, do we want to also add an `scons` architecture audit as an explicit goal? It could be that the paper would be unchanged if recompiled from scratch, but also that some `scons` dependencies are missing.

docs/production.md
```diff
@@ -127,8 +128,10 @@ Production encompasses the steps we take before we circulate or submit a paper d
 
 #### Goals
 
-  *  Main statements made in paper about definitions of samples and variables are consistent with code.
-  *  Checking every variable and sample definition can for some projects take a large amount of time. By default, this task should be limited to a single person-day of work. Unless PI notes otherwise, you should focus on checking the definition of the main sample(s) in the paper and variables in the core specification(s), and either ignore or just spot-check robustness analyses, supplemental analyses in appendices, etc.
+  *  Main statements made in paper about definitions of samples and variables, and other methodological choices, are consistent with code.
```
simonessigaberg, line 133: @jmshapir, perhaps add: "To check, verify that all non-trivial methodological statements made in the paper have a counterpart in the code, and that all non-trivial steps in the code are appropriately documented in the paper."

docs/production.md
simonessigaberg, line 161: @jmshapir, do we want to clarify scope here? I.e., 'Restrict focus to proofing the manuscript as opposed to verifying substantive claims.'

simonessigaberg, line 166: @jmshapir, for LLM integration, it could be useful to specify that the LLM can produce a list with suggested changes and comments, organized by paper section and severity.

simonessigaberg, line 181: @jmshapir, it could be helpful to specify that we also want to check for consistency in how we reference sections, tables, chapters, and the like.

simonessigaberg: @jmshapir, are 2 RAs necessary here?

simonessigaberg, line 152: @jmshapir, are 2 RAs necessary here?

simonessigaberg (changes requested):
@jmshapir, I left a few comments below.

docs/production.md
jmshapir: @simonessigaberg I like that idea--now very feasible with LLMs. See https://github.com/JMSLab/Template/commit/7c72cf5ae92baad394738d808ff46786107a9299.

docs/production.md
jmshapir: I tried making that more explicit via the deliverables in c909ff4. See what you think.

docs/production.md
jmshapir: I think (?) that's reasonably clear from the goals. Let me know if you disagree.

jmshapir: I tried to cover that somewhat generically here:
https://github.com/JMSLab/Template/blob/26de614842849461385690c79812ca676862665a/docs/production.md?plain=1#L27

I'd want to add more detail on this specific if we think that (i) there is a very likely "best" way to do this (otherwise we can leave it to discretion/experimentation) and (ii) we expect that best way to persist for some time (otherwise maintenance costs become higher).

Let me know your take on that.

jmshapir: I think we have that here?

https://github.com/JMSLab/Template/blob/26de614842849461385690c79812ca676862665a/docs/production.md?plain=1#L154

jmshapir: @simonessigaberg if this is for FACT, then I think I agree that 1LLM+1RA is probably sufficient. Changed the default in a4b0ca2.

jmshapir: For proofreading I might still suggest two RAs. Different readers tend to catch different kinds of things.

jmshapir:
@simonessigaberg thanks for helpful comments!

I replied to each. In cases where we've converged please resolve the thread; in cases where we haven't please reply!

docs/production.md
simonessigaberg: @jmshapir, that approach seems reasonable to me. I agree it is best not to assume that any 'best practices' will persist.

simonessigaberg (approved):

simonessigaberg:
@jmshapir, thanks! It looks like Github is experiencing [degraded performance](https://www.githubstatus.com/), and I am unable to resolve the threads above. We have converged on all of them and I approved the PR.

PR #161, opened 2026-05-05
PR for #160 Correct minus sign for tablefill

### LINKED ISSUE #160: Correct minus sign for tablefill
Often we have negative-valued inputs to `tablefill`. In those cases, the filled values get a hyphen instead of a proper minus sign.

Here we modify `tablefill` so that the filled values have a proper `\u2212` negative when the inputs are hyphen-negative values.

zhizhongpu:


zhizhongpu:
Interesting, maybe this doesn't work on windows. I'll think about this.

zhizhongpu:
Status: I will look into this after July 1st.

zhizhongpu:
@jmshapir I'm reasonably confident that this would be robust. Of course, I cannot foresee scenarios that I don't know, but the good news is that I'll be available for a while to maintain it.

jmshapir (changes requested):
@zhizhongpu thanks. Should we recompile (the relevant portions of) `Template` to verify that it works? If we need an example we could add, say, the difference between GDP per capita in Norway and Lichtenstein to the table.

zhizhongpu:
> Should we recompile (the relevant portions of) `Template` to verify that it works?

thanks @jmshapir. I was wondering about this. In particular, despite the SConscript specifications https://github.com/JMSLab/Template/blob/16f7a518f2199c75b05c01f06b306d6d2643eb9e/source/paper/SConscript#L3-L17 and https://github.com/JMSLab/Template/blob/16f7a518f2199c75b05c01f06b306d6d2643eb9e/source/talk/SConscript#L3-L13, `output/paper` and `output/talk` have been empty since at least two years ago. Is that intentional?

> If we need an example we could add, say, the difference between GDP per capita in Norway and Lichtenstein to the table.

Will do. Would you also like to illustrate, by the way, the use case of filling two panels of the same lyx table using two separate txt tables? A reader might find that either more useful or more complicated.

jmshapir:
> , `output/paper` and `output/talk` have been empty since at least two years ago. Is that intentional?

If you mean these omissions:

https://github.com/JMSLab/Template/blob/c69748d09c3e9fa92bb2102a9ff346f067a143cc/README.md?plain=1#L32

then yes!

> Will do. Would you also like to illustrate, by the way, the use case of filling two panels of the same lyx table using two separate txt tables? A reader might find that either more useful or more complicated.

I agree with the tradeoffs and my instinct is no.

liaochris:
@jmshapir @zhizhongpu 

In 1d3366131b37730984362e317cdcb6cf1adbec8d I 
1. added an example where the difference between two values is negative and is used as an input to Tablefill
2. Recompiled the pipeline

I think this is the last step before merging?

jmshapir (approved):

liaochris:
@zhizhongpu Let me know if we're good to merge here.

PR #162, opened 2026-05-12
Pull Request for #159: dev mode

### LINKED ISSUE #159: Dev mode
I'd like to modify the template so that by default every project has a "dev mode" that runs the build at low computational cost.

Can you propose an architecture that can live here in the Template?

Our usual goal is a MWE that illustrates the intent and implementation (but not every conceivable use case).

Thanks!

simonessigaberg:


jmshapir (commented):
@simonessigaberg this architecture works for me!

Let's get a round of input from @liaochris @zhizhongpu before we merge.

Thanks.

zhizhongpu:
I was able to replicate and both dev and full mode seem to work well.

A few reactions:

- Major:
  1. `temp/derived/wb_clean/wb_clean.json` is specified in SConscript and explicitly used in a Python script. This seems confusing from the perspective of a new user. For example:
     - Why does `temp/derived/wb_clean/wb_clean.json` look like a source node in the DAG when it's produced from some other file?
     - Why does `build.py` not load the original json file directly, and instead loads this temp json file?
     - Suppose that I first run `scons mode=dev`, but then (e.g. I realized that there's a bug somewhere) I return to development and try to run just `build.py`, I'm effectively stuck at dev mode because `temp/derived/wb_clean/wb_clean.json` is already produced and the value `20` persists. And to get out of this, I'd need to manually edit `temp/derived/wb_clean/wb_clean.json` to change the value to the one in `temp/derived/wb_clean/wb_clean.json`.
     - (This seems to be an architectural challenge and maybe your informed conclusion is that there's no more perfect / clean / intuitive solution. If so, perhaps a second best might be to document this more so that a user understands the dev mode and its risks / complications better, but let's see what @jmshapir thinks)
  2. Would it make sense to default to full mode? Isn't the expected frequency of running `dev` mode higher than `full` mode, by design? 
- Minor:
  1. Is there a way to do `scons --dev` instead of `scons --mode=dev`? Alternatively, is there a shorthand argument like how `-c` is relative to `--clean` and `-n` is to `--dry-run`? That saves a few keystrokes but the benefit accumulates over time.
  2. `json` -> `toml` to better signal that it's for humans to edit

jmshapir:
@zhizhongpu thanks!

My thoughts:
Major
(i) @simonessigaberg can you give me a short overview of why the architecture includes putting a modified JSON in TEMP? Based on that I should be able to weigh in.
(ii) If this means to say to default to dev mode, I'd probably keep the default as full.
Minor
(i) I think the call format is ok for now.
(ii) No strong views; will be curious what @simonessigaberg thinks.

simonessigaberg:
Thanks @zhizhongpu! 

@jmshapir, 

> (i) @simonessigaberg can you give me a short overview of why the architecture includes putting a modified JSON in TEMP? Based on that I should be able to weigh in.

The current architecture is structured as follows:
1. JSON in `source` specifies parameters for both modes. 
2. `SCons` creates a modified JSON in `temp` with only the parameters for the relevant build mode. 
3. Analysis scripts read the modified JSON in `temp`. Analysis scripts do not need to be aware of the build mode because the JSON in `temp` only includes the parameters for the relevant build mode. 

Here is an alternative architecture that may be clearer.
1. JSON in `source` specifies parameters for both modes. 
2. `SCons` passes the build mode to the analysis script using `CL_ARGS`.
3. Analysis script parses the JSON in `source` to extract the relevant parameters given the build mode.

My inclination is to implement the alternative architecture. Please let me know what you think @jmshapir, thanks.

> (ii) No strong views; will be curious what @simonessigaberg thinks.

I lean towards `toml` because `toml` natively supports comments and `json` does not. I can implement once we pin down the architecture.

jmshapir:
@simonessigaberg thanks and yes, I think that alternative architecture will be clearer and more robust. 

It's possible I'll want to iterate a bit on the structure. If it's not difficult to implement, maybe you can make the switch and I can take a look?

If it's difficult to implement, feel free to wireframe it so I can see how it will look and comment from there.

Thanks!

simonessigaberg:
@jmshapir, I refactored in [cc7a75e](https://github.com/JMSLab/Template/pull/162/commits/cc7a75e2d8fb9a21818ac6e243ed622d3c3455a2) and migrated to TOML in [20a239c](https://github.com/JMSLab/Template/pull/162/commits/20a239c608607664b3764b26364732fa274572f3).

We may want to create a small utility in `source/lib/` that does the following if we anticipate repeating this operation many times.

https://github.com/JMSLab/Template/blob/cc7a75e2d8fb9a21818ac6e243ed622d3c3455a2/source/derived/wb_clean/build.py#L12-L19

jmshapir:
@simonessigaberg thanks!

I agree we might want a helper like that but that also seems like something we could build later if we find we're repeating that code block a bunch.

For the base architecture, I like the revised structure better.

How hard would it be to modify the python builder so that it automatically passes the mode if specified, so that the call would just be

`env.Python(target, source)`

?

If not too difficult, that would have the advantage that the user decides this once when calling `scons`, and then only has to consider it on a script-by-script basis.

(Of course, there might be reasons to call dev details in an `SConscript` too, e.g. to specify only a subset of targets, but we wouldn't be precluding that option if we take this route.)

simonessigaberg:
@jmshapir, this is straightforward, and I implemented it in [ecba364](https://github.com/JMSLab/Template/pull/162/commits/ecba3645480c6e491f62c4ede960ed40f34e822e).

Under this architecture, the user still has to specify that `env.Value(mode)` is a source of a python build step:
https://github.com/JMSLab/Template/blob/ecba3645480c6e491f62c4ede960ed40f34e822e/source/derived/SConscript#L7-L10
Otherwise, a change in the mode would trigger a re-build of every python build step, regardless of whether the step depends on the mode. 
- This may be acceptable behavior, in which case we can automatically add `env.Value(mode)` as a source of every python build step, eliminating the need to specify the `env.Value(mode)` source.

zhizhongpu:
> @jmshapir, this is straightforward, and I implemented it in [ecba364](https://github.com/JMSLab/Template/pull/162/commits/ecba3645480c6e491f62c4ede960ed40f34e822e).

does this imply that dev mode is only supported by the Python builder, and not R, Stata, and MatLab?

simonessigaberg:
> does this imply that dev mode is only supported by the Python builder, and not R, Stata, and MatLab?

@zhizhongpu, not quite. The user is still able to pass the mode as a `CL_ARG` to an R, Stata, or MatLab builder. [ecba364](https://github.com/JMSLab/Template/pull/162/commits/ecba3645480c6e491f62c4ede960ed40f34e822e) makes this automatic only for Python.

jmshapir:
@simonessigaberg @zhizhongpu thanks both!

Catching up, my understanding is that with this architecture:
* For python: the user declares which build steps depend on `mode` by declaring `env.Value(mode)` as source for any such build step. 
* For other builders: the user passes the mode as a `CL_ARG` using the more explicit formulation we started with.

Is that right?

If so, this works for me, and I'll take another pass over the code and probably add @liaochris @zhizhongpu for more explicit review.

If not, let me know so I can better understand the flow.

simonessigaberg:
@jmshapir, 

> - For python: the user declares which build steps depend on mode by declaring env.Value(mode) as source for any such build step.
> - For other builders: the user passes the mode as a CL_ARG using the more explicit formulation we started with.

This is correct.

jmshapir (approved):

jmshapir:
(Added additional reviewers.)

README.md
```diff
@@ -72,6 +72,8 @@ In addition, each project may use other specialized tools. For the working examp
 
     - To run a given script or create a given file, run `scons path/to/script_or_file`; this will recursively run all the dependencies required to run the script or create the file.  e.g.  `scons output/derived/wb_clean/gdp_education.csv`.
 
+    - For a faster, lower-fidelity build for development, run `scons --mode=dev`. SCons passes the mode to each analysis script as a command-line argument; the script reads its TOML config and merges in the `[dev]` overrides when the mode is `dev` (see `source/derived/wb_clean/wb_clean.toml` and `source/derived/wb_clean/build.py` for an example).
```
liaochris: Phrasing "analysis script" here might be confusing because the docs are about a script in `source/derived` and "analysis" has a separate meaning in this template.

source/derived/SConscript
```diff
@@ -1,9 +1,13 @@
...
-
-target = ['#output/derived/wb_clean/gdp_education.csv']
+target = [
+    '#output/derived/wb_clean/gdp_education.csv',
+    '#output/derived/wb_clean/gdp_education.log'
+]
 source = ['#source/derived/wb_clean/build.py',
           '#source/lib/SaveData.py',
+          '#source/derived/wb_clean/wb_clean.toml',
+          env.Value(mode),
```
liaochris, line 10: @simonessigaberg 

1. How does the placement of `env.Value(mode)` determine the builder's fidelity?
2. I'm also curious about the architecture - why are we passing in `env.Value(mode)` to `source` as opposed to making it an additional argument of `env.Python` like `env.Python(target, source, env.Value(mode))`.

source/derived/wb_clean/build.py
```diff
@@ -7,7 +9,16 @@ def Main():
...
 
-    df = PrepareData(raw_dir)
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--mode", default="full")
+    mode = parser.parse_args().mode
+
+    with open("source/derived/wb_clean/wb_clean.toml", "rb") as f:
+        config = tomllib.load(f)
+    if mode == "dev":
+        config.update(config.pop("dev", {}))
```
liaochris, line 19: 1. Would it be useful to create a helper function that takes a mode and a list of settings files as inputs, then returns a config dictionary with the mode-specific settings applied?
2. I typically use all caps for config-related variables like `MODE` and `CONFIG` so they are visually distinguishable from other variables, but this is just a stylistic preference.

liaochris (changes requested):
@simonessigaberg Excited to see this come to life! I think this is a feature we will all benefit from.

source/derived/wb_clean/build.py
jmshapir: On (1), see https://github.com/JMSLab/Template/pull/162#issuecomment-4443789633.

source/derived/SConscript
simonessigaberg: @liaochris,

1. If there is a change in the mode, then a step is rebuilt if and only if `env.Value(mode)` is a source. 
2. The python builder automatically adds a CL argument with the mode. See [here](https://github.com/JMSLab/Template/blob/cc4eed4f40805663f8d55b85cd6d85c7e3a3fb36/source/lib/JMSLab/builders/build_python.py#L38-L47). It is necessary to _additionally_ specify the mode as a source per my answer to 1).

README.md
simonessigaberg: Fixed in [e5a064b](https://github.com/JMSLab/Template/pull/162/commits/e5a064bda984a0f0a58981c86465c6cd4b73a5b4).

source/derived/wb_clean/build.py
simonessigaberg: On (2), fair enough, though I am inclined to leave the variables in lower case since we use lower case [here](https://github.com/JMSLab/Template/blob/cc4eed4f40805663f8d55b85cd6d85c7e3a3fb36/source/lib/JMSLab/builders/build_python.py#L38-L47), and there is no explicit clash with the [principles](https://github.com/JMSLab/Template/blob/c69748d09c3e9fa92bb2102a9ff346f067a143cc/docs/principles.md).

simonessigaberg:
@liaochris, thanks for the comments. I have replied. Barring any objections from @jmshapir, please resolve threads where we have converged, and reply where we have not.

source/derived/SConscript
liaochris: @simonessigaberg Thanks! This makes a lot of sense and the implementation is very neat. I like that the position of `env.Value(mode)` doesn't matter (as long as it's not the first argument).

source/derived/wb_clean/build.py
liaochris: @simonessigaberg Sounds good!

liaochris (approved):

source/derived/SConscript
zhizhongpu, line 10: @simonessigaberg I love the clever solution here.

source/derived/wb_clean/build.py
```diff
@@ -7,7 +9,16 @@ def Main():
     raw_dir = Path("datastore/raw/world_bank/orig")
     out_dir = Path("output/derived/wb_clean")
 
-    df = PrepareData(raw_dir)
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--mode", default="full")
+    mode = parser.parse_args().mode
+
```
zhizhongpu, line 15: Optional: we can wrap this into a helper function so that it looks prettier 

Maybe the current way is better because if a user uses this block a lot they can also do it themselves in an instance of Template.

zhizhongpu (approved):
This looks elegant and runs well! @simonessigaberg

source/derived/wb_clean/build.py
zhizhongpu: I realized it was discussed [here](https://github.com/JMSLab/Template/pull/162#issuecomment-4443789633).

simonessigaberg:
Thanks @zhizhongpu!

@jmshapir, confirming I am set to squash and merge here?

jmshapir:
@simonessigaberg confirmed thanks!

jmshapir:
@simonessigaberg reminder to please update open branches from `main`.

simonessigaberg:
@jmshapir, 

> @simonessigaberg reminder to please update open branches from main.

Done, thanks!

PR #165, opened 2026-05-21
PR for #149: Address stata builder bugs

### LINKED ISSUE #149: stata builder bugs
Opening this issue to report two bugs with the stata builder:

1. whenever a do file contains a name with `.` in the filename before `.do`, `scons` raises an uninformative error
	- demo: 61c6ad405b19eab4066da53aafa0f2d31492fc88 
	- this only affects stata. for example, renaming `build.py` into `build.v2.py` is fine
	- expected behavior: `scons` does not raise an error, or at least raises an informative error 
2. when a do file raises an error but all outputs/targets are produced, `scons` does not raise this error 
	- demo: 6ce5c6441652afb67173eac0f6b75c84e94f9ed6
	- this only affects stata. for example, having a failed assertion at the end of a python script does induce a `scons` error
	- expected behavior: `scons` breaks

My OS is MacOS. Python version 3.12. STATA version 19.5

liaochris:
In this issue I addressed two things. 

1. `.do` files can't contain periods because stata batch mode creates logs that are named based on everything up to the first period. So `file.do` and `file.v2.do` share the log `file.log`, which creates collision errors. I fixed this by having the stata builder prohibit periods in file stems. 

2.  `scons` fails to raise errors when `.do` files fail since stata doesn't return a non-zero exit code when it fails. I fixed this by updating the scons builder to detect when `stata` builds fail. 
  - **Edit:** I also added a unit test to confirm that when stata fails, we capture the error.


jmshapir (approved):

README.md
```diff
@@ -14,7 +14,7 @@ A template for research projects developed by JMSLab.
 In addition, each project may use other specialized tools. For the working example in this template, install:
 
 - [R](https://www.r-project.org/)
-- [Stata](https://www.stata.com/install-guide/)
+- [Stata SE/MP](https://www.stata.com/install-guide/)
```
zhizhongpu, line 17: @jmshapir I added this because the current stata script `takelogs.do` never would have worked on Stata BE. With BE, [this line](https://github.com/JMSLab/Template/blob/f5dce3c9c2d978e2c63a689309f46f8ead3b4283/source/derived/wb_clean/takelogs.do#L3) runs into an error.

Note that `preliminaries` [defaults](https://github.com/gslab-econ/gslab_stata/blob/91d2c3bc0acda9334f200d769b69517cb9f3a3fe/gslab_misc/ado/preliminaries.ado#L6-L7) to `matsize 5000` and `maxvar 5000`. This is supported on the SE/MP versions which allow matsize up to 11,000, but not on BE which caps matsize at 800.

I lean towards removing the `preliminaries` package, so that set-up is easier and more people can use Template, but I also don't know much about `preliminaries` so leaving it up to you.

zhizhongpu (approved):
I tested this on the RC and my local. Everything is working as intended. 

@jmshapir let me know if you'd like to merge now or wait for @liaochris to merge.

README.md
jmshapir: @zhizhongpu on the switch to Stata SE/MP, good by me.

On `preliminaries`, I can see a case for unwinding that dependency. Are we using `gslab_misc` Stata tools anywhere else in Template?

jmshapir:
> @jmshapir let me know if you'd like to merge now or wait for @liaochris to merge.

@zhizhongpu let's wait for @liaochris to merge in case @liaochris has anything else in mind to do here first.

README.md
zhizhongpu: > On `preliminaries`, I can see a case for unwinding that dependency. Are we using `gslab_misc` Stata tools anywhere else in Template?

@jmshapir I think `preliminaries` is the only one given [this requirement](https://github.com/JMSLab/Template/blob/a68018e438381bcbfa3b118f44f153caaca972ed/source/lib/requirements.do#L1-L3) (which has been tested to be [sufficient](https://en.wikipedia.org/wiki/Necessity_and_sufficiency)).

jmshapir: Ah got it. Let's leave `preliminaries` then. It's nice to illustrate `requirements.do` and as long as we're doing that I think there's some value in pointing to that `gslab_misc` library.

liaochris:
@zhizhongpu @jmshapir fine by me to merge here, nothing else I wanted to add.

zhizhongpu:
@jmshapir fyi I just realized that the reason why GitHub Actions didn't originally create and link to an issue summary comment was that the branch name `stata-builder-bug-demo` does not start with {ISSUENUMBER}-

I added the link by editing https://github.com/JMSLab/Template/pull/165#issuecomment-4631977182

PR #167, opened 2026-06-08
PR for #139: Improving logging to avoid concurrency issues. 

### LINKED ISSUE #139: sconscript.log does not capture logs from multiple scripts
Issue demo: 4ab2405016feb04a0ab4725ca2d923a642caccda. To reproduce, go to this commit and run `scons --clean && scons`

Issue description: 

Note that under the set-up above, we have two scripts which run sequentially to produce outputs under `output/derived/wb_clean`:

- `source/derived/wb_clean/build.py`, which
	- takes more than 17 seconds
	- produces `gdp_education.csv`
	- prints "File 'output/derived/wb_clean/gdp_education.csv' saved successfully."
- `source/derived/wb_clean/takelogs.do`, which
	- takes in `gdp_education.csv`
	- takes 2 seconds
	- prints a bunch of (arguably useless) Stata logs like the below

```

  ___  ____  ____  ____  ____ ®
 /__    /   ____/   /   ____/      StataNow 19.5
___/   /   /___/   /   /___/       MP—Parallel Edition

```  

The problem is that `output/derived/wb_clean/sconscript.log` only contains the Stata print-outs. Also, the time stamps reflect only the time taken by the `.do` file. 

This has two downsides. First, a user may be interested in keeping track of warnings / print-outs from the first script; second, a user may be interested in the runtime of the first script. The current set-up does not suppose these uses.

I suspect that the underlying problem is that each script has its own `sconscript.log`, but the later one overwrites the earlier one. I think we might be able to change the builder so that `sconscript.log`s append to, instead of overwrite, each other.

fyi @jmshapir @liaochris

liaochris:
In this issue, we implemented a new logging procedure that replaces subfolder-wide`sconscript.log` files with script-specific files in a `log` folder that mirrors `source` and is not version-controlled.


output/run.log
jmshapir, line 1: Not sure we want to insist on a manifest for this file given that it is itself a kind of manifest. Maybe just assert that there is a key?

source/lib/JMSLab/builders/log.py
jmshapir, line 1: Would we want to live in `/builders`?

.github/checks/check_log_failures.py
jmshapir, line 1: Since most logs will now be invisible to this checker, do we want to ask it to check that there are no failure codes in `run.csv`?

jmshapir (changes requested):
@liaochris thanks!

A few things:
* Let's add a summary of the main changes in the lead comment of the pull.
* I made a few notes on specifics.
* If we can get a quick review from @toogiii @arewolinski that might be helpful for workflow/onboarding.

.github/checks/check_log_failures.py
liaochris: @jmshapir Great idea - makes the check much cleaner! See 5ac79a6419ed8ad719fddbf0302777da3a0bcecd

output/run.log
liaochris: Changed in 370073afb7aed2f360b401a2ed82384b681b25e0

source/lib/JMSLab/builders/log.py
liaochris: Moved in 370073afb7aed2f360b401a2ed82384b681b25e0 - I kept the name rather than renaming to `build_log.py` to distinguish it from the builder scripts that are actually `scons` environment builders.

liaochris:
@toogiii @arewolinski welcome to your first pull. Please see [reviewing principles here](https://github.com/JMSLab/Template/blob/main/docs/workflow.md).

.github/checks/check_log_failures.py
jmshapir: @liaochris thanks!

Since we're keeping `sconstruct.log`, should we check that, too?

zhizhongpu (changes requested):
Strange error in 648f6f5878d18c1528ead4d8c5a0904c6b3ea796 produced by the following:
1. `scons -c`
2. `scons`, which hangs due to matlab
3. force close terminal
4. `scons`

.github/checks/check_log_failures.py
liaochris: @jmshapir  Yes - good idea! Post- 3a2ad444277b46103630e987ca4fd4d8d16b8856, the check is now robust to various `sconstruct.log` related errors that may be caused by pipeline breakages unrelated to individual files failing (like misspecifying the DAG)!

liaochris:
@zhizhongpu 

I'm not sure what the bug described by https://github.com/JMSLab/Template/pull/167#pullrequestreview-4453905251 is. 

When I try to replicate the procedure you described in the comment, where I kill the scons run while it's working on a matlab task, I see in `output/run.csv` that the success flag for the matlab script is 0.

zhizhongpu:
> I'm not sure what the bug described by [#167 (review)](https://github.com/JMSLab/Template/pull/167#pullrequestreview-4453905251) is.

Good point @liaochris. I should have been clearer. I meant [this](https://github.com/JMSLab/Template/blob/648f6f5878d18c1528ead4d8c5a0904c6b3ea796/output/run.csv#L2-L4). I thought the key of the table is a runnable script, but here log files are documented too. Is that intended?

liaochris:
> > I'm not sure what the bug described by [(earlier review)](https://github.com/JMSLab/Template/pull/167#pullrequestreview-4453905251) is.
> 
> Good point @liaochris. I should have been clearer. I meant [this](https://github.com/JMSLab/Template/blob/648f6f5878d18c1528ead4d8c5a0904c6b3ea796/output/run.csv#L2-L4). I thought the key of the table is a runnable script, but here log files are documented too. Is that intended?

@zhizhongpu the appearance of those log files is a bug that I fixed in an earlier commit. If you replicate the test you tried earlier, can you confirm that you don't see those log files anymore? I confirmed this in my own testing.

zhizhongpu:
> > > I'm not sure what the bug described by [(earlier review)](https://github.com/JMSLab/Template/pull/167#pullrequestreview-4453905251) is.
> > 
> > 
> > Good point @liaochris. I should have been clearer. I meant [this](https://github.com/JMSLab/Template/blob/648f6f5878d18c1528ead4d8c5a0904c6b3ea796/output/run.csv#L2-L4). I thought the key of the table is a runnable script, but here log files are documented too. Is that intended?
> 
> @zhizhongpu the appearance of those log files is a bug that I fixed in an earlier commit. If you replicate the test you tried earlier, can you confirm that you don't see those log files anymore? I confirmed this in my own testing.

@liaochris  thanks

1. I can't replicate the matlab hanging issue anymore - it hangs randomly so I can't test it in a controlled environment.
2. That said, I tried to force quit the process three times during different stages of the matlab building process, and the old issue hasn't surfaced, so I think it's likely fine now and we probably don't need to worry about it.

output/talk/TemplateSlidesLatex.log
```diff
@@ -1,4 +1,4 @@
-This is pdfTeX, Version 3.141592653-2.6-1.40.26 (TeX Live 2024) (preloaded format=pdflatex 2024.7.3)  21 MAY 2026 12:18
+This is pdfTeX, Version 3.141592653-2.6-1.40.26 (TeX Live 2024) (preloaded format=pdflatex 2024.7.3)  8 JUN 2026 18:23
```
zhizhongpu: question for consideration / discussion: should this file be here or be in the git-untracked directory?

jmshapir (approved):

liaochris:
@zhizhongpu let me know what else is blocking your approval on this issue. 

@toogiii @arewolinski once you guys compile the template on `main`, as part of your pull review you should also try compiling the template on this branch.

zhizhongpu:
> @zhizhongpu let me know what else is blocking your approval on this issue.

@liaochris thanks! FYI I plan to work on this later this week.

.github/checks/check_log_failures.py
```diff
@@ -0,0 +1,33 @@
...
+RUN_CSV = Path("output") / "run.csv"
+
+def Main():
+    with open(RUN_CSV, newline="") as f:
+        rows = list(csv.DictReader(f))
+
+    script_failed = [row for row in rows if row["success"] != "1"]
+    
+    sconstruct = Path("sconstruct.log").read_text()
+    scons_failed = (
```
toogiii, line 15: are these all the possible/relevant scons build errors?

source/lib/JMSLab/builders/jmslab_builder.py
```diff
@@ -100,9 +101,25 @@ def add_log_file(self):
...
+        log_path = JMSLabBuilder.get_log_file_path(self.source_file, log_ext = log_ext)
+        os.makedirs(os.path.dirname(log_path), exist_ok = True)
+        self.log_file = log_path
+        open(log_path, 'w').close()
         return None
 
+    @staticmethod
+    def get_log_file_path(source_file, log_ext = ''):
+        '''
+        Map a source script path to its canonical per-script build log path.
```
zhizhongpu: Does the suggestion below remove important information?

```suggestion
        Map a source script path to its per-script build log path.
```

source/lib/JMSLab/tests/_test_helpers.py
```diff
@@ -63,7 +64,7 @@ def command_match(command, language, which = None):
                          r'(?P<args>.*)',
                          command)
     elif language == 'lyx':
-        # e.g. "lyx -E pdf2 target_file file.lyx > sconscript.log"
+        # e.g. "lyx -E pdf2 target_file file.lyx > log/paper/file.log"
```
zhizhongpu:
```suggestion
        # e.g. "lyx -E pdf2 target_file source/paper/file.lyx > log/paper/file.log"
```

source/lib/JMSLab/tests/_test_helpers.py
```diff
@@ -78,7 +79,7 @@ def command_match(command, language, which = None):
                          command)
 
     elif language in ['latex', 'pdflatex']:
-        # e.g. "pdflatex -interaction nonstopmode -jobname target_file file.tex > sconscript.log"
+        # e.g. "pdflatex -interaction nonstopmode -jobname target_file file.tex > log/paper/file.log"
```
zhizhongpu:
```suggestion
        # e.g. "pdflatex -interaction nonstopmode -jobname target_file source/paper/file.tex > log/paper/file.log"
```

source/lib/JMSLab/tests/_test_helpers.py
```diff
@@ -93,7 +94,7 @@ def command_match(command, language, which = None):
                          command)
 
     elif language in ['bibtex']:
-        # e.g. "bibtex target_file > sconscript.log"
+        # e.g. "bibtex target_file > log/paper/file.log"
```
zhizhongpu:
```suggestion
        # e.g. "bibtex source/paper/target_file > log/paper/file.log"
```

source/lib/JMSLab/tests/test_log.py
```diff
@@ -168,10 +185,135 @@ def test_end_log(self, mock_time):
...
 
+    def test_collect_builder_logs_from_log_directory(self):
+        builder_log = self.make_builder_log('log/input/test_script.log',
+                                            '2000-01-01 00:00:00',
+                                            '2000-01-01 00:00:01',
+                                            'succeeded')
+
+        logs = log.collect_builder_logs(TESTDIR)
+
+        self.assertIn(os.path.join('log', 'input', 'test_script.log'), logs)
```
toogiii: do you think we should add the output timestamps to these asserts (the ones in the dictionary returned by collect_builder_logs), or is that necessary? timestamp mapping did work when I compiled the template and is in the test_run_csv portion, so not convinced this is necessary.

source/lib/JMSLab/tests/test_build_stata.py
```diff
@@ -120,6 +120,27 @@ def test_bad_stata_executable(self, mock_check_output):
             build_stata(target = 'test_output.txt',
                         source = 'test_script.do',
                         env    = env)
+        helpers.check_log(self, helpers.expected_log_path('test_script.do'), 'failed')
+        original_stata_log = TESTDIR / 'test_script.log'
+        self.assertFalse(original_stata_log.exists())
+
+    @subprocess_patch
+    def test_missing_target_still_writes_log_to_log_dir(self, mock_check_output):
+        def missing_target_side_effect(*args, **kwargs):
```
zhizhongpu, line 129:
```suggestion
        def missing_target_side_effect():
```

or did I not understand this?

toogiii (commented):
left a couple comments but everything seems ok to me! 

maybe could potentially test for timestamps in [test_collect_builder_logs_from_log_directory](https://github.com/JMSLab/Template/pull/167/changes#diff-7c723392fe45d0fe6120873b9fc6e1a03860bbb5d7e4b3596de9bf8111040748R188) and subsequent tests, but this is tested in the test of run_csv implicitly later. don't think this change is necessary.

additionally, had a brief question on if the [scons_failed](https://github.com/JMSLab/Template/pull/167/changes#diff-06a157bd161958071e8804d1f8ee7799a2ea4369e96c09e886a3277742404d86R15) bool has all relevant error codes. I think it probably does, but just wanted to double check.

can approve as soon as these are answered? not totally sure how the review stack totally works yet but what's the protocol for a case like this?

output/talk/TemplateSlidesLatex.log
zhizhongpu: or should we have something like https://github.com/JMSLab/Template/blob/099bab2fbe3ff7fb8461a4c469ca674b82bf1fb1/output/paper/.gitignore#L1

? Maybe this is a question for @jmshapir

output/analysis/plots/gdp_educ.eps
```diff
@@ -359,8 +359,8 @@ end
 %%PageBoundingBox: 0 0 560 420
 %%BeginPageSetup
 N
-   45    18 M
-  555    18 L
+   45    17 M
```
zhizhongpu, line 362: My understanding is that this output changed for non-obvious mechanical reasons (e.g. machine / version differnece)?

.github/checks/check_log_failures.py
zhizhongpu: That's all I've seen during my tenure here.

zhizhongpu (approved):
Amazing @liaochris - I think this will be much better.

.github/checks/check_log_failures.py
liaochris: @toogiii I second what @zhizhongpu said; resolving here.

liaochris:
> can approve as soon as these are answered? not totally sure how the review stack totally works yet but what's the protocol for a case like this?

@toogiii protocol is to `request changes` if you have comments/changes you'd like to request, and approve once all of the threads you've opened to request changes/comment on parts of the code have been addressed.

source/lib/JMSLab/tests/test_log.py
liaochris: Might as well, see b52f9d9171ab667ef675755ba7067770d8129393. Please resolve if good.

output/analysis/plots/gdp_educ.eps
liaochris: @zhizhongpu correct! If there's nothing actionable here please resolve.

source/lib/JMSLab/tests/test_build_stata.py
liaochris: @zhizhongpu We can't do that as `mock_check_output.side_effect = missing_target_side_effect` and tge `side_effect` function takes in those two parameters as arguments.

output/talk/TemplateSlidesLatex.log
liaochris: @zhizhongpu I would advocate for [something like the latter](https://github.com/JMSLab/Template/pull/167#discussion_r3390963779); that's what we do for other projects where we use tex to compile slides.

source/lib/JMSLab/tests/_test_helpers.py
liaochris: See 8a802e324b73f40f37aedb0d31e76ef35f213308 - generalized it because LyX can also be used to compile slides.

source/lib/JMSLab/tests/_test_helpers.py
liaochris: See 8a802e324b73f40f37aedb0d31e76ef35f213308 - generalized it because TeX can also be used to compile slides.

source/lib/JMSLab/tests/_test_helpers.py
liaochris: @zhizhongpu actually, i think we should just use ur suggestion - see e2c2cc2b0ef174494d18c573cca3d6159940af5d

source/lib/JMSLab/tests/_test_helpers.py
liaochris: @zhizhongpu actually, i think we should just use ur suggestion - see e2c2cc2b0ef174494d18c573cca3d6159940af5d

source/lib/JMSLab/tests/_test_helpers.py
liaochris: @zhizhongpu I made the docstring correct to the context and I also fixed a bug where the `.log` for bibtex commands doesn't exist.

output/talk/TemplateSlidesLatex.log
jmshapir: If ffe85cb is what you have in mind then it works for me thanks!

liaochris: @zhizhongpu https://github.com/JMSLab/Template/commit/ffe85cb5c5661f387f5c8f22e25b929a183d0635 looks good with me; please resolve if you have no further comments on the thread.

source/lib/JMSLab/tests/test_log.py
toogiii: I don't have write access to the repo, so I don't think I can resolve this thread. @jmshapir would It be good for me have write access here?

jmshapir: @toogiii yes and done, please use it wisely!

output/run.csv
```diff
@@ -0,0 +1,11 @@
+filename,success,start_time,end_time
+log/paper/Template.bibtex.log,0,,
```
zhizhongpu: is this a bug?

zhizhongpu (changes requested):

output/run.csv
zhizhongpu: @liaochris I just did a local run and uploaded the outputs in 9d94a3b0bfa6f8e7add0ab266881e96f855b5c2f

we can see that https://github.com/JMSLab/Template/blob/9d94a3b0bfa6f8e7add0ab266881e96f855b5c2f/output/run.csv?plain=1#L2 got added

This seems like something that happened before. 

I wonder if there's a simple solution of just filtering out anything in $filename that does not end with `.py`, `.r`, .`do`, `.m`, `.tex`, and `.lyx`

zhizhongpu: > I wonder if there's a simple solution of just filtering out anything in $filename that does not end with `.py`, `.r`, .`do`, `.m`, `.tex`, and `.lyx`

To be clear, there should be a better way to fix this than my proposal, which can be considered as an easy / hotfix / backup

source/lib/JMSLab/tests/_test_helpers.py
zhizhongpu: @liaochris as of 18051835762d16158f4620656f75024715a3588c I'm still seeing 

```
        # e.g. "bibtex target_file > log/paper/file.log"
```
and I still suggest 
```
        # e.g. "bibtex source/paper/target_file > log/paper/file.log"
```
does your comment mean that my suggestion is undesirable?

liaochris: @zhizhongpu 

I updated the code + docstring to not have a separate bibtex log, since the bibtex log is always included in the corresponding `.tex` log when bibtex fails. Hence, including the log in `run.csv` for bibtex was an unnecessary complication on my part. 

Changes made in 36ce29b029fcb4d76ccb301eb003c136bb2d2983. I think this renders your comment moot, let me know if you have any other thoughts.

output/run.csv
liaochris: @zhizhongpu 
1. the bibtex log is gone as discussed in https://github.com/JMSLab/Template/pull/167#discussion_r3398164172
2. I can't replicate your appearance of ` log/analysis/plots/plot_gdp_educ_exp.log,0,, `, as when I compile `scons` everything works fine. How did you attain that error?

zhizhongpu (approved):
I think we're just waiting for @arewolinski to take a look.

output/run.csv
zhizhongpu: > 1. the bibtex log is gone as discussed in [PR for #139: Improving logging to avoid concurrency issues.  #167 (comment)](https://github.com/JMSLab/Template/pull/167#discussion_r3398164172)

@liaochris Great - and that seems to be the case in my local too.

> 2. I can't replicate your appearance of `log/analysis/plots/plot_gdp_educ_exp.log,0,,`, as when I compile `scons` everything works fine. How did you attain that error?

- I think that came from force-quitting my terminal to address the matlab hanging issue. 
- I can no longer replicate it after 36ce29b029fcb4d76ccb301eb003c136bb2d2983. Either 36ce29b029fcb4d76ccb301eb003c136bb2d2983 helped or it's because I can't get matlab to hang anymore.
- Since this has happened a few times, I was proposing some low-level hotfix like https://github.com/JMSLab/Template/pull/167#discussion_r3397299975. If you don't like it we can resolve it here as  it shouldn't be a huge problem if such bugs occur during only extraordinary events like a terminal apocalypse.

liaochris: ddc3265cff58cb1fd2994a069adc4cdd51ec373d

I implemented a fix in the above commit. Do you want to see if you can get matlab to hang, and if what you see aligns with what I see (which is that the matlab script in `run.csv` has a failure indicator)

zhizhongpu: thanks @liaochris. I at will make my matlab hang, but I've tried to kill the terminal a few times during a scons build, and so far nothing buggy has occurred, so perhaps 

1. we can resolve here if you are confident that ddc3265cff58cb1fd2994a069adc4cdd51ec373d is an improvement relative to the state before it
2. we can revert ddc3265cff58cb1fd2994a069adc4cdd51ec373d otherwise

zhizhongpu:
@toogiii @arewolinski I hope you can both make sure that `scons` runs well both clean (after `scons --clean`) and unclean, before you approve the pull. 

This is especially important as my machine sometimes runs into a known issue of matlab hanging

@liaochris do you think we want to execute a run on SLURM too, just in case? If so, we might as well have @arewolinski  and @toogiii do it since running the Template is typically the onboarding task for SLURM.

toogiii:
happy to try the slurm run tomorrow. the scons and scons --clean calls still work for me locally

output/run.csv
liaochris: @zhizhongpu I will leave the improvement in place.

liaochris:
> @liaochris do you think we want to execute a run on SLURM too, just in case?

@zhizhongpu works when I compile Template on slurm. 

FYI @arewolinski @toogiii you'll probably encounter some missing LaTeX `.sty` file errors when you attempt to compile on the RC. The RC has [instructions](https://docs.rc.fas.harvard.edu/kb/faq/) on how you can install LaTeX packages to overcome those errors, but if you encounter difficulty finding that documentation please let me know.

source/lib/JMSLab/tests/test_log.py
liaochris: Resolving here, FYI @toogiii

liaochris:
@toogiii @arewolinski I think your reviews are the last thing blocking this pull. Here's what I think we need to do in order to merge the pull. 

1. Both of you still need to compile the Template on the RC. This should be a similar procedure to compiling the Template on a local machine. 

2. @arewolinski I think you still need to leave review comments - please see the pull review guidelines below

https://github.com/JMSLab/Template/blob/8d79103a9d1fc238c550dde8caf9732b933d647b/docs/workflow.md#L20-L23
Compiling the template on the RC should cover point (2) above.

Let me know if you have any questions about this, and when you have done the tasks above, you can approve the pull.

arewolinski (approved):

.sconsign.dblite
toogiii, line 1: should we add this to .gitignore? would there ever be a reason to commit this

toogiii (changes requested):
only one thing @liaochris: should we add .sconsign.dblite to .gitignore? is there a reason we would want to git add this ever?

.sconsign.dblite
zhizhongpu: Good question @toogiii.

You can think of the function of this binary as tracking which processes needs to be run again after some modifications on parts of the pipeline. It is the reason why if you run `scons` immediately after a complete `scons` run, scons will say there's nothing to build.

Re: whether we should version-control it. I asked @jmshapir once and there are pros and cons. There was a discussion and the conclusion is that we git-track it.

fyi @arewolinski

toogiii (approved):
resolved other review.

zhizhongpu:
@liaochris is the SKIP here https://github.com/JMSLab/Template/pull/167#issuecomment-4752838582 expected?

PR #168, opened 2026-06-09
PR for #166: Ensure SaveData hash consistency for identical outputs

### LINKED ISSUE #166: SaveData hash inconsistency for identical outputs.
There is a bug in the `SaveData` hash logic. Right now, the dataframe hash is generated before the data are sorted and standardized, which means the same saved output can produce different hashes. For example, changing only the row order of the input to `SaveData` can change the hash even when the written output is identical.

liaochris:
Previously, there was a bug in the `SaveData` hash logic where the dataframe hash was generated before the datawas  sorted and standardized, which meant the same saved output could produce different hashes. For example, changing only the row order of the input to SaveData can change the hash even when the written output is identical.
 
 `issue` contains a script demonstrating the bug on an old `SaveData` version. 

I fixed this by moving the hash generation to after the data is standardized.


liaochris:
@jmshapir Per email, assigning @veli-m-andirin as a reviewer. LLM pr review output is in `issue/REVIEW.md`. It made one useful suggestion! 

Let me know if we want to also assign other reivewers, such as @arewolinski @toogiii

veli-m-andirin (approved):
Thanks @liaochris! Looks good to me.

jmshapir (approved):

PR #169, opened 2026-06-11
PR for #155 Improve Python Autofill

### LINKED ISSUE #155: Improve Python Autofill
@liaochris and I don't like the current implementation of `autofill.py`. Here we discuss and improve the function.

fyi @jmshapir

zhizhongpu:
https://github.com/JMSLab/Template/issues/155#issuecomment-4683006561


source/lib/JMSLab/tests/test_autofill.py
```diff
@@ -1,69 +1,74 @@
...
-        self.assertTrue("Arguments 'autofill_lists' and 'autofill_formats' are incompatible" in str(context.exception))
+    def test_append(self):
+        AutoFill({"Epsilon": -1.19}, "{:.2f}", self.outfile)
+        AutoFill({"MarginalCost": 2.59}, "{:.2f}", self.outfile, append=True)
+        with open(self.outfile) as f:
+            content = f.read()
+        self.assertEqual(content, "\\newcommand{\\Epsilon}{-1.19}\n\\newcommand{\\MarginalCost}{2.59}\n")
         return None
-        
+
```
jmshapir: Related to point (5) in https://github.com/JMSLab/Template/issues/155#issuecomment-4431710004, would we want to add a test for handling of path objects?

source/lib/JMSLab/autofill.py
```diff
@@ -1,56 +1,41 @@
...
 import inspect
+from pathlib import Path
+from typing import Any
+
+
+def AutoFill(
+    macros: dict[str, Any] | list[str],
+    format: str | None = "{:.2f}",
+    outfile: str | Path = "autofill.tex",
+    mode: str = "math",
```
jmshapir: Would something like this

```suggestion
    mode: Literal["math","text"]
```

be clearer?

jmshapir (changes requested):
In addition to minor housekeeping points, noting that I think (?) another change here is that `autofill_formats` used to to allow a list but now using a list of formats requires iterative calls. (That seems cleaner anyway but just flagging as a note.)

source/lib/JMSLab/tests/test_autofill.py
zhizhongpu: Good point. 3407fbd93d535ae146ca2ee86d0d551646bc5d3b

source/lib/JMSLab/autofill.py
zhizhongpu: good point. fixed in b77a4ba60ef1eb2a0cb1bd367bdc461c25aace40.

jmshapir (approved):

zhizhongpu:
thanks @jmshapir. I just noticed the below

> noting that I think (?) another change here is that `autofill_formats` used to to allow a list but now using a list of formats requires iterative calls.

which is a good point. In d862d55338035ff605347f257ca710a129ad9fbc I have additionally allowed specifying a list of formats when autofilling multiple macros in one go.

jmshapir:
@zhizhongpu thanks. That sounds reasonable. I also think it could be reasonable to ask the user to iterate over autofill calls (rather than to have autofill do the iteration internally), so defer to whichever architecture you prefer.

source/lib/JMSLab/autofill.py
```diff
@@ -1,56 +1,49 @@
+from __future__ import annotations
```
liaochris: Do we need this?

source/lib/JMSLab/autofill.py
```diff
@@ -1,56 +1,49 @@
...
+
+def AutoFill(
+    macros: dict[str, Any] | list[str],
+    outfile: str | Path,
+    format: str | list[str | None] | None = None,
+    append: bool = False,
+    mode: Literal["math", "text"] = "math",
+) -> None:
+    if isinstance(macros, dict):
+        resolved = macros
```
liaochris: `resolved` isn't a very clear name.

source/lib/JMSLab/autofill.py
```diff
@@ -1,56 +1,49 @@
...
+    elif isinstance(macros, list):
+        caller_frame = inspect.currentframe().f_back # type: ignore
+        resolved = {}
+        for var in macros:
+            frame = caller_frame
+            while frame is not None and var not in frame.f_locals:
+                frame = frame.f_back
+            if frame is None:
+                raise Exception(f"AutoFill: Variable '{var}' not found")
+            resolved[var] = frame.f_locals[var]
```
liaochris: I feel like we could make the code more explicit by checking, via a helper, (in order of priority)
a) for local variables in the frame
b) for global variables in the frame
c) flagging that we can't find any variables to autofill

source/analysis/top_gdp/topgdp_value.py
```diff
@@ -1,19 +1,15 @@
...
     df = pd.read_csv(instub / "gdp_education.csv")
-    TopGDPValue = df['GDP_2010'].sort_values(ascending=False).iloc[0]
+    top_gdp_value = df['GDP_2010'].sort_values(ascending=False).iloc[0]
 
-    GenerateAutofillMacros(
-        ["TopGDPValue"],
-        "{:,.0f}",
-        outstub / "top_gdp.tex"
-    )
+    AutoFill({"TopGDPValue": top_gdp_value}, outstub / "top_gdp.tex", "{:,.0f}")
```
liaochris, line 12: much better than having to define variables in the autofill name format.

source/lib/JMSLab/autofill.py
```diff
@@ -1,56 +1,49 @@
...
-    if type(autofill_formats) == str:
-        output_macros = ''.join(Autofill(autofill_var, autofill_formats) for autofill_var in autofill_lists)
+from pathlib import Path
+from typing import Any, Literal
+
+
+def AutoFill(
+    macros: dict[str, Any] | list[str],
+    outfile: str | Path,
+    format: str | list[str | None] | None = None,
```
liaochris, line 9: Do we want to explicitly enforce any format checks in `Autofill`, or just let bad formats fail via `_FormatMacro`. I don't have a strong preference for either scenario, just raising for discussion.

source/lib/JMSLab/tests/test_autofill.py
```diff
@@ -1,69 +1,95 @@
...
-from ..autofill import GenerateAutofillMacros
+from ..autofill import AutoFill
 
 class Test(TestCase):
 
     def setUp(self):
-        self.tempdir = tempfile.mkdtemp() 
-        self.outfile = self.tempdir + r"output_macros.tex"
+        self.tempdir = tempfile.mkdtemp()
+        self.outfile = self.tempdir + r"\output_macros.tex"
```
liaochris: Shouldn't we use `Path`

source/lib/JMSLab/tests/test_autofill.py
```diff
@@ -1,69 +1,95 @@
...
-        GenerateAutofillMacros([["Epsilon"], ["MarginalCost"]], 
-                               autofill_formats = ["{:.2f}", "\\textnormal{{{:.2f}}}"], 
-                               autofill_outfile = self.outfile)
-        tex_file = open(self.outfile, 'r')
-        content = tex_file.read()
-        self.assertEqual(content, "\\newcommand{\\Epsilon}{-1.19}\n\\newcommand{\\MarginalCost}{\\textnormal{2.59}}\n")
-        tex_file.close()
+    def test_list_output(self):
+        Epsilon = -1.19
+        AutoFill(["Epsilon"], self.outfile, "{:.2f}")
```
liaochris: Can we test a list with multiple values?

source/lib/JMSLab/tests/test_autofill.py
```diff
@@ -1,69 +1,95 @@
...
+        self.assertIn("Argument 'macros' must be a dict or list", str(context.exception))
+        return None
 
-        self.assertTrue("Argument 'autofill_lists' must be list" in str(context.exception))
+    def test_text_mode(self):
+        AutoFill({"MarginalCost": 2.59}, self.outfile, "{:.2f}", mode="text")
+        with open(self.outfile) as f:
+            content = f.read()
+        self.assertEqual(content, "\\newcommand{\\MarginalCost}{\\textnormal{2.59}}\n")
+        return None
```
liaochris: @zhizhongpu Should we also test math mode?

source/lib/JMSLab/tests/test_autofill.py
liaochris, line 1: @zhizhongpu two questions

1. What's the expected behavior when we try to apply a numeric format, like restricting the number of sig figs, to a string (like "ABC")?
2. Can we also test the scenario where a) the format list is incompatible with the number of entries in the autofill list, and b) the scenario where we use a list of formats?

liaochris (changes requested):

source/lib/JMSLab/autofill.py
zhizhongpu: What format checks are you thinking about? 

I don't have concern for overhead since we'll unlikely `AutoFill` 1 million times.

source/lib/JMSLab/autofill.py
zhizhongpu: 71f796b2cc19384620e4a38bc6ccbfb01ac9947d

source/lib/JMSLab/autofill.py
zhizhongpu: not sure what you mean but is it this bfdc7b8eea95ee96686027effbef16c341a02922?

if it's not, perhaps it's more efficient if you adjust it directly

source/lib/JMSLab/autofill.py
zhizhongpu: renamed in bfdc7b8eea95ee96686027effbef16c341a02922

source/lib/JMSLab/tests/test_autofill.py
zhizhongpu: > 1. What's the expected behavior when we try to apply a numeric format, like restricting the number of sig figs, to a string (like "ABC")?

I don't have a view. Ideally the user does not do that?

source/lib/JMSLab/autofill.py
liaochris: That's exactly what i meant.

source/lib/JMSLab/tests/test_autofill.py
zhizhongpu: > > 1. What's the expected behavior when we try to apply a numeric format, like restricting the number of sig figs, to a string (like "ABC")?
> 
> I don't have a view. Ideally the user does not do that?

The actual behavior is that Python's throws an `ValueError: Unknown format code 'f' for object of type 'str'` which I think is clear enough, so we don't need to do more here?

source/lib/JMSLab/autofill.py
zhizhongpu: @liaochris I rewrote the tests in a19e431b6ec608b80abbfc888fde93dddd5fc4d8

1. If that builds the format checks, please resolve
2. Else, if you don't feel strongly about having format checks, please resolve
3. Else, please suggest an example.

source/lib/JMSLab/tests/test_autofill.py
zhizhongpu: fixed in a19e431b6ec608b80abbfc888fde93dddd5fc4d8

source/lib/JMSLab/tests/test_autofill.py
zhizhongpu: fixed in a19e431b6ec608b80abbfc888fde93dddd5fc4d8

source/lib/JMSLab/tests/test_autofill.py
zhizhongpu: @liaochris math mode is tested (elsewhere) as it's the default

source/lib/JMSLab/tests/test_autofill.py
zhizhongpu: > 2\. Can we also test the scenario where a) the format list is incompatible with the number of entries in the autofill list, and b) the scenario where we use a list of formats?

Brilliant thought. Does https://github.com/JMSLab/Template/blob/a19e431b6ec608b80abbfc888fde93dddd5fc4d8/source/lib/JMSLab/autofill.py#L32-L35 work?

zhizhongpu:
@liaochris I completely rewrote the tests and it's ready for your review.

source/lib/JMSLab/tests/test_autofill.py
liaochris: > The actual behavior is that Python's throws an `ValueError: Unknown format code 'f' for object of type 'str'` which I think is clear enough, so we don't need to do more here?

This works

liaochris (approved):

PR #173, opened 2026-06-26
PR for #172 upgrade github actions checkout version

### LINKED ISSUE #172: GitHub Action check failed silently due to deprecated library
`check-issue.yml` no longer loudly fails due to the use of `actions/checkout@v4` which is no longer supported

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Here we fix this by upgrading `actions/checkout@v4` to v6.

zhizhongpu:


zhizhongpu:
@jmshapir please suggest reviewers

/run-actions-issue --post

arewolinski (approved):
From my understanding, this is an upgrade that helps us run the github actions workflow effectively. From my reading online, v6 is the new up-to-date version that we should be using, so this change makes sense! Approved!

If @jmshapir would like, I could merge this change in (I know previously @jmshapir mentioned potentially getting @zhizhongpu to merge it in, so if that would be preferred just let me know). I don't think it should be difficult since the code barely changed.

jmshapir:
@arewolinski I think it's fine for you to merge thanks.

arewolinski:
Great, I will get that done.

arewolinski:
@jmshapir I don't believe I have write permissions to the Template. Would you be able to grant these permissions so that I can merge this branch?

jmshapir:
@arewolinski done thanks!

PR #178, opened 2026-07-31
#177 fix gap in exception path for dag checker

### LINKED ISSUE #177: Fix gap in exception path for DAG checker
## Summary

`.github/checks/check_sconscripts.py` walks `source/` and flags any file that isn't referenced by path in its nearest `SConscript`, unless the file matches a pattern in `sconscript_exceptions.toml`. That exception list is only consulted by two of the checker's three code paths. The third path — files inside a subdirectory that has no `SConscript` of its own but is referenced by name in its parent's `SConscript` — never consults the exceptions file at all. Any pattern meant to except a file in such a subdirectory is silently never honored, and the checker will keep flagging it forever regardless of how the pattern is written.

## Suggested fix

Add the same `IsExcludedFile` check to the second loop, building the path the same way it's built elsewhere in the file:

```python
for f in subfiles:
    subpath = f"source/{rel}/{subdir}/{f}"
    if not IsExcludedFile(subpath) and not IsMentioned(content, f, subdir_path):
        missing_mentions.append(f"{dir_path} -> {subdir}/{f}")
```

This makes `sconscript_exceptions.toml` behave consistently regardless of which of the checker's two file-discovery code paths a given file happens to fall into.

arewolinski:
In this issue, we modified .github/checks/check_sconscripts.py such that files inside a subdirectory that have no SConscript of their own but are referenced by name in their parent's SConscript consult the exceptions file.

[LLM Review](https://github.com/user-attachments/files/30599674/issue177_check_sconscripts_review.md); no revise comments given

### DIFF
--- a/.github/checks/check_sconscripts.py
+++ b/.github/checks/check_sconscripts.py
@@ -79,7 +79,8 @@ def CollectProblems():
                 missing_dirs.append(subdir_path)
                 continue
             for f in subfiles:
-                if not IsMentioned(content, f, subdir_path):
+                subpath = f"source/{rel}/{subdir}/{f}"
+                if not IsExcludedFile(subpath) and not IsMentioned(content, f, subdir_path):
                     missing_mentions.append(f"{dir_path} -> {subdir}/{f}")
     return missing_dirs, missing_mentions
 
--- a/.github/checks/check_sconscripts.py
+++ b/.github/checks/check_sconscripts.py
@@ -81,7 +81,23 @@ def CollectProblems():
             for f in subfiles:
                 subpath = f"source/{rel}/{subdir}/{f}"
                 if not IsExcludedFile(subpath) and not IsMentioned(content, f, subdir_path):
-                    missing_mentions.append(f"{dir_path} -> {subdir}/{f}")
+        for subdir in list(dir_names):
+            subdir_path = dir_path / subdir
+            if HasLocalSConscript(subdir_path) and re.search(rf"\b{re.escape(subdir)}\b", content):
+                continue
+            for nested_dir_path, nested_dir_names, nested_file_names in os.walk(subdir_path):
+                nested_dir_path = Path(nested_dir_path)
+                nested_dir_names[:] = sorted(d for d in nested_dir_names if not IsIgnored(d))
+                nested_files = sorted(
+                    f for f in nested_file_names
+                    if not IsIgnored(f) and ShouldCheck(nested_dir_path, f)
+                )
+                for f in nested_files:
+                    nested_path = f"source/{nested_dir_path.relative_to(ROOT).as_posix()}/{f}"
+                    if not IsExcludedFile(nested_path) and not IsMentioned(content, f, nested_dir_path):
+                        missing_mentions.append(f"{dir_path} -> {nested_dir_path.relative_to(dir_path).as_posix()}/{f}")
+            if not HasLocalSConscript(subdir_path):
+                dir_names.remove(subdir)
     return missing_dirs, missing_mentions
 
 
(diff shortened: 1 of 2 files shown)


jmshapir (approved):

toogiii (commented):
do we want this subdirectory exception checking to go to an arbitrary directory depth? if not, it looks good to me!

jmshapir:
@toogiii my take on the design of this checker is that we want (i) for it to be easy for users to declare exceptions (hence the TOML), (ii) for those exceptions to be widely respected by the checker (hence this bugfix), and (iii) for execution to be fast (as with all checkers).

If there's an alternative design that better respects (ii) and doesn't worsen (iii) then I'm all for it!

.github/checks/check_sconscripts.py
```diff
@@ -79,7 +79,8 @@ def CollectProblems():
                 missing_dirs.append(subdir_path)
                 continue
             for f in subfiles:
-                if not IsMentioned(content, f, subdir_path):
+                subpath = f"source/{rel}/{subdir}/{f}"
+                if not IsExcludedFile(subpath) and not IsMentioned(content, f, subdir_path):
                     missing_mentions.append(f"{dir_path} -> {subdir}/{f}")
```
toogiii:
```suggestion
        for subdir in list(dir_names):
            subdir_path = dir_path / subdir
            if HasLocalSConscript(subdir_path) and re.search(rf"\b{re.escape(subdir)}\b", content):
                continue
            for nested_dir_path, nested_dir_names, nested_file_names in os.walk(subdir_path):
                nested_dir_path = Path(nested_dir_path)
                nested_dir_names[:] = sorted(d for d in nested_dir_names if not IsIgnored(d))
                nested_files = sorted(
                    f for f in nested_file_names
                    if not IsIgnored(f) and ShouldCheck(nested_dir_path, f)
                )
                for f in nested_files:
                    nested_path = f"source/{nested_dir_path.relative_to(ROOT).as_posix()}/{f}"
                    if not IsExcludedFile(nested_path) and not IsMentioned(content, f, nested_dir_path):
                        missing_mentions.append(f"{dir_path} -> {nested_dir_path.relative_to(dir_path).as_posix()}/{f}")
            if not HasLocalSConscript(subdir_path):
                dir_names.remove(subdir)
```

This should walk through subdirectories and apply the exceptions in nested subdirectories as well. Feel free to check my work/disregard if this seems unnecessary! Probably should test a bit more but seemed to work for a couple simple cases on my end.

toogiii (commented):
just added the subdirectory suggestion, but again totally fine if this is out of scope. This should not degrade performance, and prevent exceptions from not being applied to nested directories.

fyi @arewolinski @jmshapir

zhizhongpu:
Long time no see @arewolinski @toogiii 

Unrelated to this issue, the first external link [here](https://github.com/JMSLab/Template/blob/bde58d4576b5b085cf5adbb774280dd53d6058a6/README.md?plain=1#L108) seems broken. 

Flagging in case one of you wants to quickly fix it here. Otherwise, we can open a new issue for it.

.github/checks/check_sconscripts.py
jmshapir: If this won't meaningfully affect performance I'm happy for us to adopt. Defer to you both!

arewolinski:
@zhizhongpu 

Should be fixed here: https://github.com/JMSLab/Template/pull/178/commits/95cbd56e4a6e50f0148ac4d974fefba425c2212a :)

.github/checks/check_sconscripts.py
arewolinski: @toogiii 
Most up to date commit with this change is here: https://github.com/JMSLab/Template/pull/178/commits/c94f772463ba4173240a98e23f70b4e771cd9ca3.

toogiii (approved):

